PR #19, opened 2021-05-11
remove main-service.js. Minor change in static-asset-server.mjs

wxrdnx:
Remove main-service.js.
Some minor changes and fixes in static-asset-server.mjs:
In static-asset-server.mjs, change the static file directory to __dirname+"/../../sites" to restrict the visible page.
Also, redirect does not work with static files. Redirect only works with url. Thus, I changed "redirect" to "sendFile".
I also separated initialize() and run() in StaticAssetServer. Not sure if it is a good practice


services/main/main-server.mjs
```diff
@@ -30,13 +30,18 @@ function mainServer() {
...
+  staticAssetServer.run();
   // TODO: Uncomment the following when RPC Directory is implemented.
   // gatewayService.initialize();
 
   // TODO: Set the port once configuration is done.
   app.listen(config.get('server.port'));
 }
 
-mainServer();
+if (require.main === module) {
```
john0312, line 44: This is an ES Module, `require` is not available. You might have to try using the es-main package:
https://www.npmjs.com/package/es-main
I've tried it but it doesn't seemed to work well with yarn.

services/main/static-asset-server.mjs
```diff
@@ -31,10 +31,13 @@ class StaticAssetServer {
    */
   initialize() {
     // TODO: Restrict the visible pages.
-    this.app.use('/static', express.static(__dirname+'/../../'));
+    // Not sure if all static files are in sites
+    this.app.use('/static', express.static(__dirname+'/../../sites'));
```
john0312, line 35: We'll need common/maplib/map.mjs, which will be unavailable if we only export sites directory.

services/main/static-asset-server.mjs
```diff
@@ -31,10 +31,13 @@ class StaticAssetServer {
...
+  }
+  run() {
     // Send the user to the game client page.
     this.app.get('/', function(req, res) {
-      res.redirect('/static/sites/game-client/client.html');
```
john0312, line 37: redirect() works correctly by generating a 304 when I last tried.
If you want to send file directly, we'll lose the ability to import js/css through relative path. In that case, you might want to change client.html as well.

wxrdnx:
change from sendFile to redirect in static-asset-server.mjs

PR #21, opened 2021-05-16
Some changes in main server and static asset server

wxrdnx:
Change from sendFile to redirect in static-asset-server.mjs.
Remove `if (require.main === module) {...}` in main-server.mjs


john0312:
Merging to wip branch for fixing the git commit sequence.

PR #32, opened 2021-05-23
Add assets

john0312:


john0312:
(Added gateway service commit to this PR because of dependency.)

canyugs (approved):

PR #37, opened 2021-05-23
Merge map constructor

john0312:


john0312:
Note: Only the last commit is the one changed.

ktpss95112 (approved):

PR #46, opened 2021-05-29
Add Mock Redis

s3131212:


common/rpc-directory/mock-redis.mjs
```diff
@@ -0,0 +1,75 @@
...
+
+    get(key, cb) {
+        if (this._data.has(key)){
+            cb(null, this._data.get(key));
+        } else {
+            cb(null, null); // If the key does not exist the special value nil is returned.
+        }
+    }
+
+    set(key, value, cb) {
```
john0312, line 37: We currently use the form set([key, value], cb).

john0312: and we need to support 'NX' attribute.

common/rpc-directory/mock-redis.mjs
```diff
@@ -0,0 +1,75 @@
...
+
+    on(event, callback) {
+        if (event == 'message') {
+            this._mockRedis.callbacks.push(callback);
+        }
+    }
+
+    hget(key, cb) {
+        if (this._data.has(key)){
+            cb(null, this._data.get(key));
```
john0312, line 64: hget/hset works a level deeper.
i.e. cb(null, this._data.get(key).get(field))

john0312 (approved):

common/rpc-directory/mock-redis.mjs
s3131212: Resolved.

common/rpc-directory/mock-redis.mjs
s3131212: Resolved. I also changed the signature of `hget` and `hset`, please check if those are correct.

john0312:
Still LGTM.

PR #57, opened 2021-06-19
basic collision check

whyang9701:
簡單實作 多玩家不重疊 / 碰撞偵測

### DIFF
--- a/services/gateway/gateway-service.mjs
+++ b/services/gateway/gateway-service.mjs
@@ -204,13 +204,13 @@ class GatewayService {
     //overspeed , are you fly?
     if(distanceSquire > speed ^ 2){
       this.broadcastResetUser(socket,uid,positionA.x,positionA.y,positionA.facing)
-
       return
     }
 
     //enter none empty grid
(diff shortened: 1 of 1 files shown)


services/gateway/gateway-service.mjs
```diff
@@ -178,15 +180,43 @@ class GatewayService {
...
+      return;
+    }
+
+    // near by grid check  
+    if(!this._nearByGridCheck(msg,lastRecord)){
+      return ;
+    }
+
+    // obstacle check
+    let players = this.broadcaster.gameState.getPlayers();
```
john0312, line 207: getPlayers() can be slow because it does a fullscan with redis.

services/gateway/gateway-service.mjs
```diff
@@ -178,15 +180,43 @@ class GatewayService {
...
+    let playerIDs = Object.keys(players);
+    let obstacle = playerIDs.find( playerID => players[playerID].x === msg.x && players[playerID].y === msg.y)
```
john0312, line 209: `players` might not have all the players because we can have multiple gateway servers.

PR #68, opened 2021-07-04
extension "chat": private message done

zeze-zeze:


extensions/chat/client.mjs
```diff
@@ -18,17 +18,38 @@ class Client {
   }
 
   send_msg(client){
-    return function(evt){
+    return async function(evt){
```
john0312, line 21: Note/Optional: If you use arrow function, `this` binding will not be overridden, and you can use this in your function instead of passing `client` in.

extensions/chat/standalone.mjs
```diff
@@ -24,9 +23,16 @@ class Standalone {
...
+  async c2s_broadcastMessage(player, arg){
     // TODO: Check whether message is command like /help
-    this.helper.broadcastToAllUser(arg);
+    arg['msg_from'] = player.playerID;
+    await this.helper.broadcastToAllUser(arg);
+  }
+
+  async c2s_privateMessage(player, arg){
+    arg['msg_from'] = player.playerID;
+    await this.helper.callS2cAPI(arg.msg_to, 'chat', 'getPrivateMessage', 5000, arg);
```
john0312, line 34: Optional: You can do both at the same time with Promise.all().
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all
```
await Promise.all([this.helper.callS2cAPI(...), this.helper.callS2cAPI(...)])
```

john0312 (approved):

PR #69, opened 2021-07-04
Finish feature: multimap

ktpss95112:


common/maplib/map.mjs
```diff
@@ -7,7 +7,159 @@
...
+ *
+ *           ...
+ *
+ *   (2,0)   (2,1)   (2,2)
+ *
+ *   (1,0)   (1,1)   (1,2)   ...
+ *
+ *   (0,0)   (0,1)   (0,2)
+ */
+class MapCoord {
```
john0312, line 26: [Optional] Consider moving MapCoord into another .mjs for better compatibility with import().

extensions/jitsi/client.mjs
```diff
@@ -81,7 +81,7 @@ class Client {
   onSelfLocationUpdated(loc) {
     const map = this.helper.getMap();
     if (typeof map === 'object') {
-      let m = map.getCell('jitsi', loc.x, loc.y);
+      let m = map.getCell('jitsi', loc.mapCoord.x, loc.mapCoord.y);
```
john0312, line 84: Maybe `map.getCell(loc.mapCoord, 'jitsi')`?

sites/game-client/map-renderer.mjs
```diff
@@ -66,19 +68,22 @@ class MapRenderer {
...
-   * @param {Number} x - The x coordinate (map coordinate).
-   * @param {Number} y - The y coordinate (map coordinate).
+   * @param {MapCoord} coord - The map coordinate.
+   * @param {Number} dx - Delta x to be added. Default is 0.
+   * @param {Number} dy - Delta y to be added. Default is 0.
    */
-  setViewerPosition(x, y) {
-    const mapSize = this.map.getMapSize();
+  setViewerPosition(coord, dx, dy) {
+    dx ??= 0;
```
john0312, line 76: [Optional] I think some team members are still on older version of nodejs that doesn't support this. We should tell them to upgrade.

john0312 (approved):

common/maplib/map.mjs
ktpss95112: I think that it is reasonable to place this class in map.mjs, since it is a class related to map.
Besides, this module uses named export to export this class and uses default export for `GameMap`. This is for compatibility with the existing code.

extensions/jitsi/client.mjs
ktpss95112: You are right.
I will fix this in next PR.

sites/game-client/map-renderer.mjs
ktpss95112: I think Allen has used this syntax in previous commits. If everyone is okay with that, then there should be no concern here.

PR #81, opened 2021-07-11
Add multi-process mode and standalone extensions

s3131212:
1. Add multi-process mode, which can be enabled by `"multiprocess": true` in the config file.
2. Add standalone extension. Use `./run-ext helloworld` to run a standalone extension.
3. Add documentation about how multi-process mode and standalone API work. Maybe someday I (or someone else) will finish the documentation.


common/rpc-directory/multi-process-RPC-directory.mjs
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * Method for responing gRPC, should only be called by gRPC server.
+   * @param {Object} call - The request content, contains the same signature as callRPC.
+   * @param {Function} callback - Callback function.
+   * @param {Object} handlers - The handler of this class.
+   */
+  async responseGrpcCall(call, callback, callRPC) {
+    callback(null, {
+      response: JSON.stringify(await callRPC.bind(this)(call.request.callerServiceName, call.request.serviceName, call.request.methodName, ...JSON.parse(call.request.args)))
```
john0312, line 74: [Optional] We could consider having a callLocalRPC() and callRemoteRPC().
responseGrpcCall() would only call callLocalRPC(), callRPC() would call both. This way if there's any mismatch in server configuration, we'll not result in an infinite loop to ourself.

services/gateway/all-area-broadcaster.mjs
```diff
@@ -81,30 +80,30 @@ class AllAreaBroadcaster {
...
-   * @param {object} loc - The location object, see Gateway Service for doc.
+   * Register onLocation function.
+   * @param {Function} callback - This is called when we've a location
+   * message from redis.
    */
-  onLocation(loc) {
-    // Broadcast the location message.
-    this.io.emit('location', loc);
+  registerOnLocation(callback) {
+    this.onLocation = callback;
```
john0312, line 88: You might need to do `this.onLocation = () => {};` in the constructor to make sure even if registerOnLocation() is not called, we don't crash.

Same goes to others below.

john0312 (approved):

PR #85, opened 2021-07-17
feature:  basic dialog extension

DieYiSu:


extensions/dialog/client.mjs
```diff
@@ -0,0 +1,73 @@
...
+    let arg = {'npc': 'Amy'} //for test
+
+    this.helper.callC2sAPI(null, 'getSpecificNpcData', 5000, arg);
+  }
+
+  /**
+   * Show the single choice dialog to user.
+   * @param {object} arg - npc data.
+   */
+  s2c_singleChoiceDialog(arg){
```
john0312, line 34: These should be async, and will timeout (close the dialog) or return when the user does something.

extensions/dialog/client.mjs
```diff
@@ -0,0 +1,73 @@
...
+    document.getElementById('npcSentence').innerHTML = arg.sentence;
+    document.getElementById('checkButtonDiv').innerHTML = buttonString;
+    document.getElementById('checkButtonDiv').addEventListener('click', e => {this.sendResultToServer(e.target.innerHTML)});
+  }
+
+  /**
+   * Send the user choice to Server.
+   * @param result - user choice.
+   */
+  sendResultToServer(result){
```
john0312, line 67: This should be part of the 2 calls above, or resolves the Promise from these 2 calls.

extensions/dialog/standalone.mjs
```diff
@@ -0,0 +1,82 @@
...
+    this.npcObj = JSON.parse(await fs.promises.readFile(filepath, 'utf8'));
+  }
+
+  /**
+   * Get the specific NPC data from Server (npc.json), 
+   * than send NPC data back to user to render the dialog.
+   * @param {object} player - Player ID
+   * @@param {object} arg - NPC name
+   */
+  async c2s_getSpecificNpcData(player, arg){
```
john0312, line 42: We'll discuss where do we implement NPC functionalities in the meeting.

extensions/dialog/client.mjs
```diff
@@ -0,0 +1,73 @@
...
+   * Show the single choice dialog to user.
+   * @param {object} arg - npc data.
+   */
+  s2c_singleChoiceDialog(arg){
+    console.log(`[Dialog] Single Choice Dialog`);
+
+    document.getElementById('npcName').innerHTML = arg.name;
+    document.getElementById('npcSentence').innerHTML = arg.sentence;
+
+    //todo: show the pop-up dialog for user
```
john0312, line 40: Maybe add timeout?

extensions/dialog/client.mjs
```diff
@@ -0,0 +1,73 @@
...
+
+
+  /**
+   * Show the multi choice dialog to user.
+   * @param {object} arg - npc data.
+   */
+  s2c_multiChoiceDialog(arg){
+    console.log(`[Dialog] Multi Choice Dialog`);
+    let buttonString = '';
+    for(let choice of arg.choice){
```
john0312, line 51: Maybe have an ID for each choice? (Instead of just having a display text.)

PR #86, opened 2021-07-17
Modify coordinate checking to support multi world and add Teleport() in Gateway

john0312:


services/gateway/gateway-service.mjs
```diff
@@ -181,14 +187,13 @@ class GatewayService {
...
-    firstLocation.x = socket.playerData.x;
-    firstLocation.y = socket.playerData.y;
+    firstLocation.mapCoord = socket.playerData.mapCoord;
     firstLocation.facing = 'D';
     firstLocation.displayChar = socket.playerData.displayChar;
     socket.playerData.lastMovingTime = Date.now();
-    //try occupy grid
-    await this.dir.redis.setAsync(
-      [`${firstLocation.x}-${firstLocation.y}`, 'occupied', 'NX']);
+    // try occupy grid
```
ktpss95112, line 194: try **_occupying_** grid

services/gateway/gateway-service.mjs
```diff
@@ -181,14 +187,13 @@ class GatewayService {
...
+    // TODO: Handle cases whereby we can't occupy the location.
```
ktpss95112, line 195: The use of **_whereby_** here is awkward... do you mean _**when**_?

services/gateway/gateway-service.mjs
```diff
@@ -181,14 +187,13 @@ class GatewayService {
...
+    await this._occupyCoord(firstLocation.mapCoord, playerID);
```
ktpss95112, line 196: I think for now we should use a while loop to guard this function call instead of let it go.
```
while (! await this._occupyCoord(firstLocation.mapCoord, playerID));
```
Imagine that there is no spawn point that is not occupied. We should not let the player spawn since it might mess something later.

services/gateway/gateway-service.mjs
```diff
@@ -245,53 +250,84 @@ class GatewayService {
...
     // TODO: Check if facing is valid.
     // TODO: Check if movement is legal.
 
-    let mapSize = this.gameMap.getMapSize();
-    let lastMsg = { x:socket.playerData.x , y:socket.playerData.y}
-    if(lastMsg.x === undefined || lastMsg.y === undefined){
-      lastMsg.x = 0;
-      lastMsg.y = 0;
+    let lastCoord = socket.playerData.mapCoord;
+    let mapSize = this.gameMap.getMapSize(lastCoord.mapName);
```
ktpss95112, line 264: Use `const` whenever possible. (maybe you forgot to run `eslint`?)

services/gateway/gateway-service.mjs
```diff
@@ -245,53 +250,84 @@ class GatewayService {
...
+    if(lastCoord.x === undefined || lastCoord.y === undefined || lastCoord.mapName === undefined){
+      // Shouldn't happen, log an error.
+      console.error(`Invalid lastCoord in onUserLocation ${lastCoord}`);
+      return;
     }
     if(!this._movementRequestSpeedCheck(socket.playerData)){
+      console.warn(`Player ${msg.playerID} is speeding`);
```
ktpss95112, line 271: "Overspeeds" is more clear than " is speeding".

services/gateway/gateway-service.mjs
```diff
@@ -245,53 +250,84 @@ class GatewayService {
...
-    let lastMsg = { x:socket.playerData.x , y:socket.playerData.y}
-    if(lastMsg.x === undefined || lastMsg.y === undefined){
-      lastMsg.x = 0;
-      lastMsg.y = 0;
+    let lastCoord = socket.playerData.mapCoord;
+    let mapSize = this.gameMap.getMapSize(lastCoord.mapName);
+    if(lastCoord.x === undefined || lastCoord.y === undefined || lastCoord.mapName === undefined){
+      // Shouldn't happen, log an error.
+      console.error(`Invalid lastCoord in onUserLocation ${lastCoord}`);
+      return;
```
ktpss95112, line 268: Do we really need this check?

services/gateway/gateway-service.mjs
```diff
@@ -245,53 +250,84 @@ class GatewayService {
...
+        `${lastCoord.mapName} to ${msg.mapCoord.mapName} without permission.`);
       return;
     }
     // target position is in the map
-    if(!this._borderCheck(msg,mapSize)){
-      // restrict user position
-      msg.x = lastMsg.x;
-      msg.y = lastMsg.y;
-      await this._broadcastUserLocation(msg);
+    if(!this._borderCheck(msg.mapCoord,mapSize)){
```
ktpss95112, line 280: `eslint` will say you have bad spacing in this line.

services/gateway/gateway-service.mjs
```diff
@@ -310,24 +346,22 @@ class GatewayService {
...
       return false;
     }
     return true;
   }
 
-  _nearByGridCheck(msg,lastRecord){
-    let lastPosition = {x:lastRecord.x , y:lastRecord.y};
-    let targetPosition = {x:msg.x , y:msg.y};
-    let distanceSquire = this._getDistanceSquare(lastPosition,targetPosition);
+  _nearByGridCheck(coord,lastCoord){
```
ktpss95112, line 363: _**Nearby**_ is a word, so maybe it should be `_nearbyGridCheck`

services/gateway/gateway-service.mjs
```diff
@@ -245,53 +250,84 @@ class GatewayService {
...
-      // restrict user position
-      msg.x = lastMsg.x;
-      msg.y = lastMsg.y;
-      await this._broadcastUserLocation(msg);
+    if(!this._borderCheck(msg.mapCoord,mapSize)){
+      console.warn(`Player ${msg.playerID} is trying to go outside the map.`);
       return;
     }
-
     // near by grid check
```
ktpss95112, line 284: _**Nearby**_ is a word.

services/gateway/gateway-service.mjs
```diff
@@ -245,53 +250,84 @@ class GatewayService {
...
       return;
     }
     // target position is in the map
-    if(!this._borderCheck(msg,mapSize)){
-      // restrict user position
-      msg.x = lastMsg.x;
-      msg.y = lastMsg.y;
-      await this._broadcastUserLocation(msg);
+    if(!this._borderCheck(msg.mapCoord,mapSize)){
+      console.warn(`Player ${msg.playerID} is trying to go outside the map.`);
```
ktpss95112, line 281: Should we place the player right before the border instead of canceling their move?

services/gateway/gateway-service.mjs
```diff
@@ -245,53 +250,84 @@ class GatewayService {
...
-      // restrict user position
-      msg.x = lastMsg.x;
-      msg.y = lastMsg.y;
-      await this._broadcastUserLocation(msg);
+    if(!this._nearByGridCheck(msg.mapCoord,lastCoord)){
+      console.warn(`Player ${msg.playerID} is trynig to teleport.`);
       return;
     }
     
+    // TODO: Maybe log any cell collision.
```
ktpss95112, line 290: Do we really need this logging?

services/gateway/gateway-service.mjs
```diff
@@ -245,53 +250,84 @@ class GatewayService {
...
-    //try occupy grid
-    let ret = await this.dir.redis.setAsync(
-      [`${msg.x}-${msg.y}`, 'occupied', 'NX']);
-    // grid has be occupied
-    if(ret === null){
-      // restrict user position
-      msg.x = lastMsg.x;
-      msg.y = lastMsg.y;
-      await this._broadcastUserLocation(msg);
+    if(!this._nearByGridCheck(msg.mapCoord,lastCoord)){
```
ktpss95112, line 285: `eslint`

services/gateway/gateway-service.mjs
ktpss95112: I don't see the need for this function. Isn't the functionality already been included in the speed check?

services/gateway/gateway-service.mjs
```diff
@@ -245,53 +250,84 @@ class GatewayService {
...
+      console.error(`Invalid lastCoord in onUserLocation ${lastCoord}`);
+      return;
     }
     if(!this._movementRequestSpeedCheck(socket.playerData)){
+      console.warn(`Player ${msg.playerID} is speeding`);
+      return;
+    }
+    if (!(lastCoord.mapName === msg.mapCoord.mapName)) {
+      console.warn(`Player ${msg.playerID} changed map from ` +
+        `${lastCoord.mapName} to ${msg.mapCoord.mapName} without permission.`);
```
ktpss95112, line 276: I think this check should be included in the speed check.

services/gateway/gateway-service.mjs
```diff
@@ -245,53 +250,84 @@ class GatewayService {
...
-      msg.x = lastMsg.x;
-      msg.y = lastMsg.y;
-      await this._broadcastUserLocation(msg);
+    if(!this._nearByGridCheck(msg.mapCoord,lastCoord)){
+      console.warn(`Player ${msg.playerID} is trynig to teleport.`);
       return;
     }
     
+    // TODO: Maybe log any cell collision.
+    await this._teleportPlayerInternal(socket, msg);
```
ktpss95112, line 291: Why call `_teleportPlayerInternal` instead of `teleportPlayer`?

services/gateway/gateway-service.mjs
```diff
@@ -310,24 +346,22 @@ class GatewayService {
     return true;
   }
 
-  _getDistanceSquare(a,b){
-    return Math.pow(Math.abs(a.x - b.x),2) + Math.pow(Math.abs(a.y - b.y),2);
+  _getManhattanDistance(a,b){
```
ktpss95112, line 349: I think this should be a utility function in `map.mjs` instead of a member function of `GatewayService`.

services/gateway/gateway-service.mjs
```diff
@@ -310,24 +346,22 @@ class GatewayService {
...
+    return Math.abs(a.x - b.x) + Math.abs(a.y - b.y);
   }
 
-  _borderCheck(msg,mapSize){
-    if(msg.x > mapSize.width || msg.x < 0){
+  _borderCheck(coord,mapSize){
+    if(coord.x > mapSize.width || coord.x < 0){
```
ktpss95112, line 354: Add jsdoc, please.
The available range of x coordinate is [0, mapSize.width), so the check should be
```
if (coord.x >= mapSize.width || ...)
```

services/gateway/gateway-service.mjs
```diff
@@ -310,24 +346,22 @@ class GatewayService {
...
       return false;
     }
-    if(msg.y > mapSize.height || msg.y < 0){
+    if(coord.y > mapSize.height || coord.y < 0){
```
ktpss95112, line 357: `>=` instead of `>`

services/gateway/gateway-service.mjs
```diff
@@ -339,11 +373,54 @@ class GatewayService {
       console.log('player has no last time record')
       return false;
     }
-    if( Date.now() - playerData.lastMovingTime < moveingRequestThreshold){
+    if( Date.now() - playerData.lastMovingTime < movingRequestThreshold){
       return false;
     }
     return true;
   }
+
```
ktpss95112, line 381: Where is the obstacle check? You should block players if they try to walk across a wall.

services/gateway/gateway-service.mjs
```diff
@@ -339,11 +373,54 @@ class GatewayService {
...
+    if (ret === null) {
+      // Failed
+      return false;
+    }
+    if (ret === 'OK') {
+      return true;
+    }
+    console.assert(`Invalid reply from redis for _occupyCoord: ${ret}`);
+    return false;
+  }
```
ktpss95112, line 400: Is `getRedis().setAsync()` going to return any value other than `OK` and `null`? I think you should find the documentation to ensure what value it is going to return.
```
const ret = await ...();
return (ret === 'OK');
```

services/gateway/gateway-service.mjs
```diff
@@ -339,11 +373,54 @@ class GatewayService {
...
+
+  /**
+   * Clear the cell specified by mapCoord.
+   * If playerID is not undefined or null, then verify that it was occupied.
+   */
+  async _clearOccupy(mapCoord, playerID) {
+    if (!(playerID === null || playerID === undefined)) {
```
ktpss95112, line 407: I think `if (playerID) {` is sufficient in this case.

services/gateway/gateway-service.mjs
```diff
@@ -339,11 +373,54 @@ class GatewayService {
...
+      // Do the check
+      let getRet = await this.dir.getRedis().getAsync(
+        [mapCoord.toRedisKey()]);
+      if (!(getRet === playerID)) {
```
ktpss95112, line 411: `if (getRet !== playerID) {`

ktpss95112 (changes requested):
I think that these game-logic-related methods/functions should be inside `GameState` class. It is quite counterintuitive that we should look up `gateway-service.mjs` to find the implementation of moving a player. `gateway-service.mjs` should act as an interface only.
If I misunderstood something, please leave comments to let me know.

Also, add jsdoc, please. Run `eslint`, please.

services/gateway/gateway-service.mjs
john0312: Lisa's PR that's coming up soon will resolve this, so I would prefer leaving this here.

services/gateway/gateway-service.mjs
john0312: This is a guard so that we'll know when we screw up playerData.

services/gateway/gateway-service.mjs
john0312: This is needed because speed check checks only the time, this checks the distance.

services/gateway/gateway-service.mjs
john0312: I thought the two is the same?

services/gateway/gateway-service.mjs
john0312: It should happen very rarely in actual production. If it happens too frequently it'll point to something that we should pay attention to.

services/gateway/gateway-service.mjs
john0312: Will leave this to whyang when he moves everything that can be moved into the library.

services/gateway/gateway-service.mjs
john0312: Because we already have the socket and message.

services/gateway/gateway-service.mjs
john0312: I'll leave this to whyang.

services/gateway/gateway-service.mjs
john0312: I'll leave the task of adding jsdoc to whyang.

services/gateway/gateway-service.mjs
john0312: I'll leave this to whyang.

services/gateway/gateway-service.mjs
john0312: Generally it doesn't. The assertion is there to make sure it doesn't, and if it does due to any reasons, we'll know.

john0312:
The purpose of this change is just to get whyang's previous PR to a state whereby it doesn't block the development. As discussed in the meeting, whyang will move the logics into GameState or another class in maplib. The task of adding jsdoc and fixing eslint will also be left to whyang.

services/gateway/gateway-service.mjs
ktpss95112: Not exactly. They are the same only if the player's speed is less than 1 cell per tick.
If the player can move at, say, 2 cells per tick, and the distance between him and the border is 1 cell, he would not able to move if we cancel his move. However, I would expect that he can move 1 cell (that is, right in front of the wall).

services/gateway/gateway-service.mjs
ktpss95112: I think we should use `assert` instead of `console.error` if our intention is to know when we screw up something.
```
assert('x' in lastCoord && 'y' in lastCoord && 'mapName' in lastCoord);
```

ktpss95112: I am wondering that `socket.playerData.mapCoord` is guaranteed to be a `class MapCoord` in `GatewayService.addSocket`. I mean, a crash of the server is reasonable if we really mess up something. We should not handle this error if this is not allowed to happen.

services/gateway/gateway-service.mjs
ktpss95112: The definition of speed is distance divided by time. The speed check should check whether a player **_moves how far_** within **_how much time_**.

services/gateway/gateway-service.mjs
ktpss95112: Isn't the documentation of Redis lists all possible return values?

ktpss95112:
I understand that these changes are not complete features yet. However, I hope this merge is still robust and intuitive so that the code in the future will be in the correct direction.

john0312:
For the task that this aims to unblock (Spawn Point), I don't think it'll affect the correct direction.
As for other future work, those will commence only after whyang fixes the stuffs you outlined, so that's not an issue either.

Do we have a consensus that we want to unblock spawn point implementation and leave the other stuff to whyang, or do you want to have a meeting to discuss this?

ktpss95112 (approved):

ktpss95112:
OK, let's merge this PR.

PR #87, opened 2021-07-19
Add basic structure for the MainUI and Overlay

john0312:


fanlan1210 (approved):

common/extlib/client-extension-manager.mjs
```diff
@@ -90,7 +92,6 @@ class ClientExtensionManager {
    * @param {String} extName - The name of the extension.
    */
   async startExtensionClient(extName) {
-    // TODO: Call gameStart() on each of the extensions.
     if (!typeof extName === 'string') {
```
ktpss95112, line 95:
```
typeof extName !== 'string'
```
The current code evaluates to `(! typeof extName) === 'string'`

sites/game-client/client.ejs
```diff
@@ -22,10 +22,33 @@ absolute URL.
...
-        <%- include(i); %>
-      <% }); %>
-      <canvas id="mapcanvas"></canvas>
+      <div id="staging-div">
+        <!--
+        This div is for stuff that's not yet being displayed.
+        It is currently semi-transparent for debugging purpose.
+        -->
+        <div id="map-render-overlay" class="container-center">
+          <canvas id="mapcanvas" width="640" height="640"></canvas>
```
ktpss95112, line 31: The size of canvas is adjusted in `misc.js`. Maybe shouldn't assign the value in html?

sites/game-client/misc.js
```diff
@@ -6,7 +6,7 @@ function onResize(event) {
   const canvas = document.querySelector('#mapcanvas');
   // TODO: more detailed tweak
   canvas.height = 32 * 15; // 15 tiles
-  canvas.width = 32 * 9; // 9 tiles
+  canvas.width = 32 * 20; // 20 tiles
```
ktpss95112, line 9: I think we can move `misc.js` into `MainUI`. `misc.js` is designed to do the job of main UI previously.
Maybe create `_onResize` as a private method of `MainUI` and `registerOnResize` to let extension register their callback. `window.addEventListener` in `MainUI`'s constructor.

ktpss95112: Add TODO to `MainUI` if this is to be resolved in future development.

sites/game-client/ui/main-ui.mjs
```diff
@@ -43,28 +54,80 @@ class MainUI {
...
   // resizeTopbarOverlay(count);
+
+  // ========== Internal Methods ============
+  // Methods below should only be used by UI classes such as Model/Overlay.
+
+  /**
+   * Send a DOM element back into staging area.
+   */
+  restoreToStaging(dom) {
+    document.getElementById(STAGING_ID).appendChild(dom);
```
ktpss95112, line 93: If `STAGING_ID` is a constant, consider caching the DOM element instead of querying it every time.

ktpss95112 (commented):
Reviewed and left some comments.

sites/game-client/client.ejs
john0312: Sorry, forgot to remove when debugging.

sites/game-client/ui/main-ui.mjs
```diff
@@ -43,28 +64,90 @@ class MainUI {
...
+  }
 
   // If we want to use individual overlay for meeting
   // position = 0...n
   // setTopbarOverlay(position, overlay);
   // count=0 for no topbar.
   // resizeTopbarOverlay(count);
+
+  // ========== Internal Methods ============
+  // Methods below should only be used by UI classes such as Model/Overlay.
```
ktpss95112, line 97: Maybe start below function with an underscore as a programmer's hint, since not everyone reads this comment.

sites/game-client/ui/main-ui.mjs
```diff
@@ -43,28 +64,90 @@ class MainUI {
...
+    return true;
+  }
+
+  /**
+   * This is called whenever the window is resized.
+   * @private
+   */
+  _onResize(event) {
+    for (const overlay of this.overlays.values()) {
+      overlay.onResize(event);
```
ktpss95112, line 148: Maybe
a) generate a warning when an overlay does not implement `onResize()` in development environment
or
b) force every overlay to implement `onResize()`

This code is going to crash if an overlay does not implement `onResize()`.

sites/game-client/ui/overlay.mjs
```diff
@@ -1,113 +1,158 @@
...
-    -> HIDDEN -> SHOW -+
-       |  ^-------|    |
-       +---------------+
-       V
-    DISMISSED
-    */
+  /**
+   * If true, user can hide the overlay.
+   * 
+   * This property should be implemented by the derived class.
```
ktpss95112, line 22: Consider adding `console.error(function_name + 'should be implemented in derived class!')` to ensure the implementation.

ktpss95112 (approved):

sites/game-client/ui/main-ui.mjs
john0312: That was not done because JS doesn't have friend class so marking them private may cause issues with some linters. However, if you've concerns we'll go that way. Underscore added.

sites/game-client/ui/main-ui.mjs
john0312: There's an empty Overlay.onResize(), so it won't crash. Derived classes may/can implement onResize() but it's not strictly required if doing nothing on onResize() event is intended. Comment is added/adjusted to reflect this fact.

sites/game-client/ui/overlay.mjs
john0312: The comment is the one that's incorrect. I've adjusted the comment to reflect the fact that implementing that method is optional. i.e. "can" vs "should".

ktpss95112 (approved):

PR #89, opened 2021-07-21
Enable extension to draw its customized layer

ktpss95112:
Refactor `MapRenderer` to support a more flexible rendering mechanism.


ktpss95112:
I have fixed GameState yesterday and force-pushed it.

john0312 (approved):

PR #91, opened 2021-07-22
Teleport feature done

zeze-zeze:
1. Modify extension helper to call teleport function of RPC
2. Add chat command

### DIFF
--- a/extensions/chat/client.mjs
+++ b/extensions/chat/client.mjs
@@ -76,7 +76,7 @@ Usage: !/<Command> <arg1> <arg2> ...
    * @handleCommand
    * @param {string} cmd - the command including arguments
    */
-  handleCommand(cmd){
+  async handleCommand(cmd){
     if(cmd === '!/help'){
       this.listCommand();
     }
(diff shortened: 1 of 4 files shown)


extensions/chat/client.mjs
```diff
@@ -40,33 +40,103 @@ class Client {
...
+      if(isNaN(parseInt(mapCoord.x, 10)) || isNaN(parseInt(mapCoord.y, 10))){
+        document.getElementById('message_history').innerHTML += '<span>Invalid Coordinate</span><br>';
+        this.listCommand();
+      }
+      else{
+        this.helper.callC2sAPI(null, 'teleport', 5000, mapCoord);
+      }
+    }
+    else{
+      const result = await Promise.resolve(this.helper.callC2sAPI(null, 'adminCommand', 5000));
```
john0312, line 97: [Optional] Maybe call it otherCommands or hiddenCommands? We can have hidden commands for escape room as well.

john0312 (approved):

extensions/chat/client.mjs
zeze-zeze: Done. Change to otherCommands

john0312 (approved):

PR #94, opened 2021-07-25
Create a class CellSet to provide a unified interface for managing cell sets

ktpss95112:
In this PR, I refactor the code related to cell sets.
* Create a class `CellSet` to provide a unified interface for cell sets
* Fix `cellsettest` extension so that it provides a correct example of using dynamic cell sets

I think this PR may affect some teammates' work. If so, maybe we can merge this PR before the next meeting.

### DIFF
--- a/TODO
+++ b/TODO
@@ -19,7 +19,3 @@
 
 - maplib
   - move game-state.mjs out
-  - create map-utils.mjs
-    - class MapCoord
-    - class CellSet
-    - utility function: tells if a MapCoord is inside a CellSet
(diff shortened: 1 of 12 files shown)


john0312 (approved):

john0312:
Looks fine to me. If we're officially moving to a "cellset belongs to a single map" model, then we might need some methods for retrieving cell sets with the same name in the entire GameMap. We'll discuss in the next meeting.

s3131212 (approved):

PR #99, opened 2021-07-31
Add scripts to start and maintain all services 

s3131212:
**WIP, do not merge.**

1. Add a script `start-all.mjs` to start all gateway services and standalone extensions.
2. Maintain the healthcheck of all services and kill all if one is dead.
3. Don't initiate standalone extensions in gateway services in multi-process mode. (WIP)


john0312:
Recent progress looks good, you might need to:
- Create a separate GatewayServer class that starts a GatewayService.
- Have a mechanism to start multiple GatewayServer.
- Optionally have a class that runs the AssetServer in its own process.

services/assets/asset-server.mjs
```diff
@@ -0,0 +1,100 @@
+// Copyright 2021 HITCON Online Contributors
```
john0312, line 1: There are recent changes in asset-service.mjs that was not picked up by this PR.

I would recommend not swapping asset-service and asset-server in this PR, because that can make it hard to review. As for the final state, I'm OK whether you swap asset-service and asset-server or not.

services/main/asset-server.mjs
```diff
@@ -1,97 +1,50 @@
...
+  //   console.error(err);
+  // });
+
+  /* Create the http service */
+  const app = express();
+  const server = http.createServer(app);
+
+  /* Create services */
+  const extensionManager = new ExtensionManager(null, null, null, null);
+  console.log(Object.values(config.get('gatewayServers')).map(v => v.httpAddress));
```
john0312, line 35: Do you meant to remove this?

PR #101, opened 2021-07-31
Spawn Point

lisasasasa:
Rebase to PR #94 and Change Spawn Point Arrays' type.

### DIFF
--- a/common/maplib/map.mjs
+++ b/common/maplib/map.mjs
@@ -226,7 +226,7 @@ class GameMap {
    * Similar to getOriginalCellSet(), but matches cell set name's as prefix.
    * @param {String} csName - The prefix of cell set name.
    * @return {Map} ret - Key: mapName; Value: list of cell set.
-  **/
+   */
   getOriginalCellSetStartWith(csName) {
     const ret = new Map();
     for (const [mapName, map] of this._maps) {
@@ -237,14 +237,22 @@ class GameMap {
   }
 
   /**
-   * Get the SpawnPointList
-   * @return {mapCoord Array} : All the spawn points
+   * Get the Spawn Points from all maps.
+   * @return {mapCoord Array} : Random spawn points with mapCoord type
   **/
-  getSpawnPoint() {
+  getRandomSpawnPoint() {
     if (this.spawnPointList.length === 0) {
       this._maps.forEach((map) => {
         this.spawnPointList.concat(map.getSpawnPoint());
       });
+    }
+
+    // Suffle the spawn points array
+    for (let i = this.spawnPointList.length - 1; i > 0; i--) {
+      const j = Math.floor(Math.random() * (i + 1));
+      const temp = this.spawnPointList[i];
+      this.spawnPointList[i] = this.spawnPointList[j];
+      this.spawnPointList[j] = temp;
     }
     return this.spawnPointList;
   }
@@ -276,8 +284,6 @@ class _SingleGameMap {
       cellSet.dynamic = false;
       this.gameMap.cellSets[i] = CellSet.fromObject(cellSet);
     }
-    this.spawnpointCellSet = [];
-    this.spawn_point_list = [];
 
     // initialize dynamic cell set
     this.dynamicCellSet = {};
@@ -455,32 +461,25 @@ class _SingleGameMap {
   }
 
   /**
-   * Expand Cell Set List.
-   * @param {string} cellSetName
-   * @return {list} Cell Set points - A list of cell set points.
+   * Get spawn points of single map.
+   * If We have a spawn point area from (1,1) (1,5) (5,1) (5,5),
+   * we split it here to become 16 spawn points.
+   * @return {MapCoord Array} Spawn points - A list of spawn points with mapcoord type.
    */
-  expandCellSet(cellSetName) {
+  getSpawnPoint() {
     const expandList = [];
-    if (this.staticCellSet.has(cellSetName)) {
-      for (const cell in this.staticCellSet.get(cellSetName)) {
+    if (this.staticCellSet.has('spawnPoint')) {
+      for (const cell in this.staticCellSet.get('spawnPoint')) {
         const w = (cell.w ?? 1);
         const h = (cell.h ?? 1);
         for (let i = 0; i < w; ++i) {
           for (let j = 0; j < h; ++j) {
-            expandList.push(new MapCoord(this.gameMap, cell.x + i , cell.y + j));
+            expandList.push(new MapCoord(this.gameMap, cell.x + i, cell.y + j));
           }
         }
       }
     }
     return expandList;
-  }
-
-  /**
-   * Get Spawn Point List.
-   * @return {list} spawn point - A list of the spawn point.
-   */
-  getSpawnPoint() {
-    return this.expandCellSet('SpawnPoint');
   }
 }
 
--- a/common/maplib/map.mjs
+++ b/common/maplib/map.mjs
@@ -123,7 +123,6 @@ class GameMap {
   constructor(asset, maps) {
     this.graphicAsset = asset;
     this._maps = new Map();
-    this.spawn_point_list = [];
     for (const [mapName, map] of Object.entries(maps)) {
       this._maps.set(mapName, new _SingleGameMap(asset, map));
     }
@@ -133,7 +132,7 @@ class GameMap {
    * Remove all the dynamic cell set.
    */
   removeAllDynamicCellSet() {
-    for (const map of this._maps.values()) {
+    for (const map of this._maps) {
       map.removeAllDynamicCellSet();
     }
   }
@@ -220,23 +219,13 @@ class GameMap {
       if (cs) ret.set(mapName, cs);
     }
     return ret;
-  /**
-   * Get Spawn Point List.
-   * @param {MapCoord} coord - The map coordinate.
-   * @return {list} spawn point - A list of the spawn point.
-   **/
-  getSpawnPoint(coord){
-    if(this.spawn_point_list.length == 0){
-      this.spawn_point_list = this._maps.get(coord.mapName).getSpawnPoint(coord.x, coord.y);
-    }
-    return this.spawn_point_list;
   }
-
   /**
    * Similar to getOriginalCellSet(), but matches cell set name's as prefix.
    * @param {String} csName - The prefix of cell set name.
    * @return {Map} ret - Key: mapName; Value: list of cell set.
-   */
+  **/
+
   getOriginalCellSetStartWith(csName) {
     const ret = new Map();
     for (const [mapName, map] of this._maps) {
@@ -245,13 +234,16 @@ class GameMap {
     }
     return ret;
   }
-}
+
+  getSpawnPoint(mapName){
+    return this._maps.get(mapName).getSpawnPoint();
+  }
 
 
 /**
  * The class that represents a single map.
  * The exported GameMap is composed of many _SingleGameMap.
- */
+ **/
 class _SingleGameMap {
   /**
    * Construct an empty map.
@@ -451,38 +443,36 @@ class _SingleGameMap {
     if (tile === null) return null;
     return this.graphicAsset.getTile(layer, tile);
   }
+
   /**
-   * Get Spawn Point List.
-   * @param {int} movex , movey
-   * @return {list} spawn point - A list of the spawn point.
+   * Expand Cell Set List.
+   * @param {string} cellSetName
+   * @return {list} Cell Set points - A list of cell set points.
    */
-<<<<<<< HEAD
-  getSpawnPoint(movex, movey){
-    if(this.spawn_point_list.length == 0){
-      for(cell in this.spawnpointCellSet){
-        let w = (cell.w ?? 1), h = (cell.h ?? 1);
-        for( i in range(w)){
-          for( j in range(h)){
-            this.spawnpointCellSet.push({x: movex + cell.x + i, y:  movey + cell.y + j})
-          }
+  expandCellSet(cellSetName) {
+    let expandList = [];
+    if(this.staticCellSet.has(cellSetName))
+    for (const cell in this.staticCellSet.get(cellSetName)) {
+      const w = (cell.w ?? 1), h = (cell.h ?? 1);
+      for (let i = 0; i < w; ++i) {
+        for (let j = 0; j < h; ++j) {
+          expandList.push({x: cell.x + i, y: cell.y + j});
         }
       }
     }
-    return this.spawn_point_list;
-=======
-  getSpawnPoint(){
-    let spawn_point_list = [];
-    for(i in range(this.gameMap.width)){
-      for(j in range(this.gameMap.height)){
-        if(getCell({i, j}, "ground") == "P"){
-          spawn_point_list.push({x: i, y: j});
-        }
-      }
-    }
-    return spawn_point_list;
->>>>>>> Choose spawnpoint list
+    return expandList;
   }
 
+  /**
+   * Get Spawn Point List.
+   * @return {list} spawn point - A list of the spawn point.
+   */
+  getSpawnPoint(){
+    if (this.spawnPointList == 0) {
+      this.spawnPointList = this.expandCellSet("SpawnPoint");
+    }
+    return this.spawnPointList;
+  }
 }
 
 export default GameMap;
--- a/common/maplib/map.mjs
+++ b/common/maplib/map.mjs
@@ -479,8 +479,12 @@ class _SingleGameMap {
    */
   getSpawnPoints() {
     if (typeof this._spawnPointList === 'undefined') {
-      if (this.staticCellSet.has('spawnPoint')) {
-        for (const cell in this.staticCellSet.get('spawnPoint')) {
+      const spawnPointsCellSet = getOriginalCellSet('spawnPoint');
+      if (typeof spawnPointsCellSet === 'undefined') {
+        this._spawnPointList = [];
+      }
+      else{
+        for (const cell in spawnPointsCellSet.cells) {
           const w = (cell.w ?? 1);
           const h = (cell.h ?? 1);
           for (let i = 0; i < w; ++i) {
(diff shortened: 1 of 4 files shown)


john0312 (approved):

s3131212 (changes requested):

common/maplib/map.mjs
```diff
@@ -438,6 +453,35 @@ class _SingleGameMap {
     if (tile === null) return null;
     return this.graphicAsset.getTile(layer, tile);
   }
+
+  /**
+   * Expand Cell Set List.
+   * @param {string} cellSetName
+   * @return {list} Cell Set points - A list of cell set points.
```
ktpss95112, line 460: There is no `list` in JavaScript. Maybe you mean `Array`?

common/maplib/map.mjs
```diff
@@ -438,6 +453,35 @@ class _SingleGameMap {
...
+   */
+  expandCellSet(cellSetName) {
+    const expandList = [];
+    if (this.staticCellSet.has(cellSetName)) {
+      for (const cell in this.staticCellSet.get(cellSetName)) {
+        const w = (cell.w ?? 1);
+        const h = (cell.h ?? 1);
+        for (let i = 0; i < w; ++i) {
+          for (let j = 0; j < h; ++j) {
```
ktpss95112, line 469: Use `const`.

common/maplib/map.mjs
```diff
@@ -438,6 +453,35 @@ class _SingleGameMap {
...
+          }
+        }
+      }
+    }
+    return expandList;
+  }
+
+  /**
+   * Get Spawn Point List.
+   * @return {list} spawn point - A list of the spawn point.
```
ktpss95112, line 480: Replace `list` with `Array`.

common/rpc-directory/directory.mjs
```diff
@@ -293,8 +293,7 @@ class Directory {
 
     // TODO: Set them to a spawn point specified by map.json.
     // {MapCoord} mapCoord - Character's map coordinate.
-    setDefaultValue(playerData, 'mapCoord', new MapCoord('world1', 10, 10));
-
+    setDefaultValue(playerData, 'mapCoord', undefined);
```
ktpss95112, line 296: Why not initialize the mapCoord here? Why initialize it in `GatewayService`?

services/gateway/gateway-service.mjs
```diff
@@ -185,15 +185,34 @@ class GatewayService {
...
+      while (setLocation === false && index < spawnPoint.length) {
+        firstLocation.mapCoord = spawnPoint[index];
+        setLocation = await this._occupyCoord(firstLocation.mapCoord, playerID);
+        index = index + 1;
+      }
+      if (setLocation === false) {
+        console.warn(`No free spawn point.`);
+        socket.disconnect();
+        return;
+      }
```
ktpss95112, line 209: You had better add comments if other people cannot get what the code does at first glance.

services/gateway/gateway-service.mjs
```diff
@@ -185,15 +185,34 @@ class GatewayService {
...
+    }
     firstLocation.facing = 'D';
     firstLocation.displayChar = socket.playerData.displayChar;
     socket.playerData.lastMovingTime = Date.now();
     // try occupying grid
     // TODO: Handle cases when we can't occupy the location.
-    await this._occupyCoord(firstLocation.mapCoord, playerID);
```
ktpss95112, line 196: Why delete this line?

common/maplib/map.mjs
```diff
@@ -234,13 +235,25 @@ class GameMap {
...
+  **/
+  getSpawnPoint() {
+    if (this.spawnPointList.length === 0) {
+      this._maps.forEach((map) => {
+        this.spawnPointList.concat(map.getSpawnPoint());
+      });
+    }
+    return this.spawnPointList;
+  }
+}
```
ktpss95112, line 251: I think you should use something like `this.getOriginalCellSet('spawnPoint')`.

common/maplib/map.mjs
```diff
@@ -263,6 +276,8 @@ class _SingleGameMap {
       cellSet.dynamic = false;
       this.gameMap.cellSets[i] = CellSet.fromObject(cellSet);
     }
+    this.spawnpointCellSet = [];
+    this.spawn_point_list = [];
```
ktpss95112, line 280: 1. CamelCaseNamingConvention is commonly used in JavaScript.
2. I think there is no need to add these two class attributes. See the comment of my review.

common/maplib/map.mjs
```diff
@@ -438,6 +453,35 @@ class _SingleGameMap {
...
+        const h = (cell.h ?? 1);
+        for (let i = 0; i < w; ++i) {
+          for (let j = 0; j < h; ++j) {
+            expandList.push(new MapCoord(this.gameMap, cell.x + i , cell.y + j));
+          }
+        }
+      }
+    }
+    return expandList;
+  }
```
ktpss95112, line 476: It is hard to understand what the function does from the function name and the comment. What does "expand" mean in this context?
By the way, I don't see the need for this function. See the comments of my review.

ktpss95112 (changes requested):
In my opinion, we can initialize players' location and deal with spawn points more elegantly.
* `GameMap` has a method `getAllSpawnPoints`, which uses `getOriginalCellSet('spawnPoint')` to implement. You should handle the cache (`this._allSpawnPointCache`) in this function instead of initialize it in the constructor.
* `GameMap` has a method `getRandomSpawnPoint`, which uses `getAllSpawnPoints` to implement. You might want to declare a function locally like `randomChoice` to extract a random element of an array.
* I think we should handle spawn points as a static cell set. If so, there is no need to change `_SingleGameMap`. The existing functionality can handle static cell set very well.
* I think that initializing players' first location in `setDefaultValue` is straightforward. Why separate the functionality to `GatewayService`?

common/maplib/map.mjs
lisasasasa: If I have a spawn point area from (1,1) (1,5) (5,1) (5,5), I want to split it here to become 16 spawn points. Normally, getSpawnPoint() will only be called one time.

common/rpc-directory/directory.mjs
lisasasasa: We don't know which spawn point is empty, so we should use this._occupyCoord(firstLocation.mapCoord, playerID) to find it.

services/gateway/gateway-service.mjs
lisasasasa: Add some comments.

common/maplib/map.mjs
```diff
@@ -234,13 +235,33 @@ class GameMap {
     }
     return ret;
   }
-}
 
+  /**
+   * Get the Spawn Points from all maps.
+   * @return {mapCoord Array} : Random spawn points with mapCoord type
```
ktpss95112, line 241: Just `{Array}`, not `mapCoord Array`.

ktpss95112, line 241: No, this function should return a single spawn point.
`@return {MapCoord} A random spawn point`

common/maplib/map.mjs
```diff
@@ -234,13 +235,33 @@ class GameMap {
...
+  **/
+  getRandomSpawnPoint() {
```
ktpss95112, line 243: This function should return **a random spawn point** instead of an array of spawn points.

common/maplib/map.mjs
```diff
@@ -438,6 +459,28 @@ class _SingleGameMap {
     if (tile === null) return null;
     return this.graphicAsset.getTile(layer, tile);
   }
+
+  /**
+   * Get spawn points of single map.
+   * If We have a spawn point area from (1,1) (1,5) (5,1) (5,5),
+   * we split it here to become 16 spawn points.
+   * @return {MapCoord Array} Spawn points - A list of spawn points with mapcoord type.
```
ktpss95112, line 467: Just {Array}, not mapCoord Array.
`@return {Array} The map coordinates inside 'spawnPoint' cell set.`

common/maplib/map.mjs
```diff
@@ -438,6 +470,29 @@ class _SingleGameMap {
     if (tile === null) return null;
     return this.graphicAsset.getTile(layer, tile);
   }
+
+  /**
+   * Get spawn points of single map.
+   * If We have a spawn point area from (1,1) (1,5) (5,1) (5,5),
+   * we split it here to become 16 spawn points.
+   * @return {Array} The map coordinates inside 'spawnPoint' cell set.
+   */
```
Sea-n, line 479: This PR contains 3 types of [JSDoc](https://jsdoc.app/about-block-inline-tags.html).

```js
/**
 * First
**/
 ```

```js
/**
 * First
 **/
 ```

```js
/**
 * First
 */
 ```

File: services/gateway/gateway-service.mjs [hunk 14]
```diff
@@ -185,15 +185,21 @@ class GatewayService {
...
       displayName: socket.playerData.displayName};
     firstLocation.mapCoord = socket.playerData.mapCoord;
+    if (firstLocation.mapCoord === undefined) {
+      // Try to find the empty spawn points.
+      while (setLocation === false) {
+        firstLocation.mapCoord = this.gameMap.getRandomSpawnPointNoStarvation();
+        setLocation = await this._occupyCoord(firstLocation.mapCoord, playerID); // try occupying grid
+        if (setLocation) break;
+      }
+    }
```
Sea-n, line 198: It will cause an error.

```
ReferenceError: setLocation is not defined
```

s3131212 (approved):

common/maplib/map.mjs
```diff
@@ -234,13 +235,44 @@ class GameMap {
...
+      this._spawnPointCurrentIndex = 0; // prevent starvation
+    }
+
+    this._spawnPointCurrentIndex += 1;
+    if (this._spawnPointCurrentIndex === this._spawnPointCache.length) {
+      randomSuffle(this._spawnPointCache);
+      this._spawnPointCurrentIndex = 0;
+    }
+    return this._spawnPointCache[this._spawnPointCurrentIndex];
+  }
```
Sea-n, line 269: This statement will cause an error after fixed the previous one.

```
file:///..../hitcon-online/common/maplib/map.mjs:257
        this._spawnPointCache.concat(map.getSpawnPoints());
                              ^

TypeError: Cannot read property 'concat' of undefined
    at file:///..../hitcon-online/common/maplib/map.mjs:257:31
    at Map.forEach (<anonymous>)
    at GameMap.getRandomSpawnPointNoStarvation (file:///..../hitcon-online/common/maplib/map.mjs:256:18)
    at GatewayService.addSocket (file:///..../hitcon-online/services/gateway/gateway-service.mjs:194:47)
```

Sea-n (changes requested):
This code can't run on my machine, could you test it again?

common/maplib/map.mjs
```diff
@@ -438,6 +470,29 @@ class _SingleGameMap {
...
+        for (const cell in this.staticCellSet.get('spawnPoint')) {
+          const w = (cell.w ?? 1);
+          const h = (cell.h ?? 1);
+          for (let i = 0; i < w; ++i) {
+            for (let j = 0; j < h; ++j) {
+              this._spawnPointList.push(new MapCoord(this.gameMap, cell.x + i, cell.y + j));
+            }
+          }
+        }
+      }
```
Sea-n, line 492: Is this error expected? I didn't see where the `staticCellSet` variable was declared.

```
Player yKSkP connected.
file:///..../hitcon-online/common/maplib/map.mjs:483
      if (this.staticCellSet.has('spawnPoint')) {
                             ^

TypeError: Cannot read property 'has' of undefined
    at _SingleGameMap.getSpawnPoints (file:///..../hitcon-online/common/maplib/map.mjs:483:30)
    at file:///..../hitcon-online/common/maplib/map.mjs:258:42
    at Map.forEach (<anonymous>)
    at GameMap.getRandomSpawnPointNoStarvation (file:///..../hitcon-online/common/maplib/map.mjs:257:18)
    at GatewayService.addSocket (file:///..../hitcon-online/services/gateway/gateway-service.mjs:194:47)
```

lisasasasa: Yes, I forgot I rebase last time. Wait me a minute !

services/gateway/gateway-service.mjs
john0312: _occupyCoord is called above, so we don't need to call it again.

services/gateway/gateway-service.mjs
```diff
@@ -185,15 +185,22 @@ class GatewayService {
     }
 
     // Synchronize the state.
-    let firstLocation = {playerID: playerID, 
+    const firstLocation = {playerID: playerID,
       displayName: socket.playerData.displayName};
     firstLocation.mapCoord = socket.playerData.mapCoord;
+    let setLocation = false;
+    if (firstLocation.mapCoord === undefined) {
```
john0312, line 192: I would strongly recommend always running the spawn point assignment. To spawning existing user into where they previously left, we'll need to occupy the cell, and if that fails we'll need more algorithms. Thus, for now we should default to assigning spawn points for all new connections.

john0312:
I agree that we should cache the points with `getRandomSpawnPoint()`.

As for moving the default values for spawn point out of `setDefaultValue()`, it makes sense because:
- `RPCDirectory` doesn't know which cell is occupied and that functionality is in `GatewayService`. Thus, if we do assign a position in `RPCDirectory`, it might be occupied.
- `RPCDirectory` doesn't know about the `GameMap`, and it is intentional that we want to separate the the `GameMap` logic from `RPCDirectory`. Assigning a static `MapCoord` value could result in something that isn't available on the current map.
- It is a bonus to be able to distinguish between someone who had a position (previously logged in), vs someone who is new (undefined position).

As for `GameMap.getRandomSpawnPoint()`, I think it doesn't work as well because the logic to occupy cell resides outside the `GameMap` so `GameMap` can't return something that is free to occupy. In that case, `GatewayService` will have to call `GameMap.getRandomSpawnPoint()` multiple times to get something that it can successfully occupy. Such a scenario would call for a `getRandomSpawnPoint()` implementation that could return non-repeating sequence of spawn points and a way to indicate such a sequence have been exhausted, such an implementation would require state to be stored outside of `GameMap`, complicating the implementation and user, and as such I would recommend that it would be simpler for `GatewayService` to acquire the entire sequence of spawn points instead of having a `getRandomSpawnPoint()`.

services/gateway/gateway-service.mjs
ktpss95112:
```javascript
while (true) {
    firstLocation.mapCoord = this.gameMap.getRandomSpawnPointNoStarvation();
    if (await this._occupyCoord(firstLocation.mapCoord, playerID)) break;
}
```

How about this?

common/maplib/map.mjs
```diff
@@ -234,13 +235,44 @@ class GameMap {
...
-}
 
+  /**
+   * Get the Spawn Points from all maps without starvation.
+   * @return {mapCoord} : A random spawn point with mapCoord type
+  **/
+  getRandomSpawnPointNoStarvation() {
+    function randomSuffle(arr) {
+      // Suffle the spawn points array
+      for (let i = arr.length - 1; i > 0; i--) {
```
ktpss95112, line 246: Use const.

common/maplib/map.mjs
```diff
@@ -234,13 +235,44 @@ class GameMap {
...
+        const j = Math.floor(Math.random() * (i + 1));
+        const temp = arr[i];
+        arr[i] = arr[j];
+        arr[j] = temp;
+      }
+    }
+
+    if ( typeof this._spawnPointCache === 'undefined') {
```
ktpss95112, line 254: Redundant space.

common/maplib/map.mjs
```diff
@@ -234,13 +235,44 @@ class GameMap {
...
+      this._maps.forEach((map) => {
+        this._spawnPointCache.concat(map.getSpawnPoints());
+      });
+      randomSuffle(this._spawnPointCache);
+      this._spawnPointCurrentIndex = 0; // prevent starvation
+    }
+
+    this._spawnPointCurrentIndex += 1;
+    if (this._spawnPointCurrentIndex === this._spawnPointCache.length) {
+      randomSuffle(this._spawnPointCache);
```
ktpss95112, line 265: No need shuffle here.

common/maplib/map.mjs
```diff
@@ -438,6 +470,33 @@ class _SingleGameMap {
...
+      const spawnPointsCellSet = getOriginalCellSet('spawnPoint');
+      if (typeof spawnPointsCellSet === 'undefined') {
+        this._spawnPointList = [];
+      }
+      else{
+        for (const cell in spawnPointsCellSet.cells) {
+          const w = (cell.w ?? 1);
+          const h = (cell.h ?? 1);
+          for (let i = 0; i < w; ++i) {
+            for (let j = 0; j < h; ++j) {
```
ktpss95112, line 491: Use const.

services/gateway/gateway-service.mjs
ktpss95112: I agree with this. I suggest adding a TODO to remind you of finishing this feature in the future.

ktpss95112:
If @lisasasasa implements `getRandomSpawnPoint` correctly, the function should return non-repeating random spawn points in sequence. I think that offering an API for getting the whole list of spawn points is okay, but using `getRandomSpawnPoint` will simplify gateway-service significantly. If there is currently no spawn point available, it is reasonable that the spawning process just busy-wait for an available spawn point. Obtaining the whole list of spawn points and implement this logic in gateway-service is overkill in my opinion.

```
while (true) {
    firstLocation.mapCoord = this.gameMap.getRandomSpawnPointNoStarvation();
    if (await this._occupyCoord(firstLocation.mapCoord, playerID)) break;
}
```

ktpss95112 (approved):

john0312:
In the next meeting, we'll discuss if we want to disconnect the user when there's no available spots or we'll busy loop until there are available spots. Meanwhile, note that we can have multiple new connections coming in concurrently, and they may call `getRandomSpawnPointNoStarvation()` in an interleaved manner.

PR #84, opened 2021-07-16
完成 items extension 中 standalone server 的相關函式。

wxrdnx:
完成 items extension 中 standalone server 的相關函式。目前需要思考的事情是如何在主畫面呈現道具，以及相關道具的動畫效果應該要如何去達成。

PR #106, opened 2021-08-11
Fix the problem that PR #99 accidentally removed code from PR #84, and GameState not synchronizing correctly

s3131212:
1. PR #99 accidentally remove the changes to asset server from PR #84 , this PR adds the changes back.
2. In multiprocess mode, GameState in the gateway services should be updated when received updates from Redis.

### DIFF
--- a/run/start-all.mjs
+++ b/run/start-all.mjs
@@ -60,7 +60,20 @@ async function main() {
     gatewayServers[serverName] = fork('../services/main/gateway-server.mjs', ['--service-name', serverName], { cwd: '.' });
   }
 
-  await new Promise(r => setTimeout(r, 4000)); // TODO: better way to prevent race condition.
+  // Wait for all gateway services to start.
+  const promises = Object.keys(gatewayServers).map((serverName) => {
+    return new Promise((resolve, reject) => {
+      gatewayServers[serverName].on('message', msg => {
+        if(msg !== 'started') return;
+        console.log(`${serverName} has started successfully.`)
+        resolve(msg)
+      });
+      const timer = setTimeout(() => {
+        reject(new Error(`Promise timed out after 10 sec.`));
+      }, 10 * 1000);
+    });
+  });
+  await Promise.all(promises);
 
   /* Start standalone extension services */
   const extServices = {};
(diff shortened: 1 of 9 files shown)


run/start-all.mjs
```diff
@@ -60,11 +60,12 @@ async function main() {
     gatewayServers[serverName] = fork('../services/main/gateway-server.mjs', ['--service-name', serverName], { cwd: '.' });
   }
 
+  await new Promise(r => setTimeout(r, 4000)); // TODO: better way to prevent race condition.
```
s3131212, line 63: By the way, I think I need some help on this part. How do I ensure that standalone extension are executed **after** gateway services are fully started?
In some cases, a standalone extension might publish to Redis channel before gateway services are subscribe to that channel, causing the GameState to be out of sync.

ktpss95112: I think that in JS you should use an `async` function `launchGatewayService(someService)` to handle the startup of a service. In such way, you can do this:
```javascript
const allServices = [...];
const promises = allServices.forEach((s) => {return launchGatewayService(s)});
await Promise.all(promises);
```

I also saw you use `fork` to launch the gateway server. You may need a ping-pong message to know if your gateway server is ready.

run/start-all.mjs
```diff
@@ -60,11 +60,25 @@ async function main() {
...
+      gatewayServers[serverName].on('message', msg => {
+        if(msg !== 'started') return;
+        console.log(`${serverName} has started successfully.`)
+        resolve(msg)
+      });
+      const timer = setTimeout(() => {
+        reject(new Error(`Promise timed out after 10 sec.`));
+      }, 10 * 1000);
+    });
+  });
```
ktpss95112, line 75:
```javascript
const primises = Object.entries(gatewayServers).map((entry) => {
    const [name, server] = entry;
    server.on(...);
    ...
})

```

ktpss95112: Error message should be more informative. For example, `Failed to start "${name}" server successfully in ${timeout} seconds.`.

ktpss95112 (approved):

PR #108, opened 2021-08-13
Add Docker, Nginx load balancing, and CORS validation

s3131212:
1. Now we can start hitcon-online using docker-compose.
2. Add Nginx for reverse proxy and load balancing.
3. Add CORS validation if the public address is known.
4. Update documentation.

## Instructions
Update the hosts file.
```
127.0.0.1       online.hitcon.org
127.0.0.1       gateway.online.hitcon.org
```

Start docker-compose 
```
cd docker
docker-compose up
```

Access the front-end at `online.hitcon.org` .


docker/nginx-template.conf
```diff
@@ -0,0 +1,37 @@
...
+        default upgrade;
+        '' close;
+    }}
+
+    upstream gateways {{
+        ip_hash;
+        {gateways}
+    }}
+
+    server {{
```
john0312, line 19: Optional: For ease of deployment, I would recommend having only 1 virtual host, that way we can make it the default, and possibly do away with cross origin entirely. The advantage is that if someone wants to spin up an instance of hitcon-online, there would be no need to setup multiple DNS before doing docker compose up. We do have only 1 virtual host by having different location point to different upstream. i.e. location /static and location /client.html goes to asset service.

docker/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+
+  redis:
+    image: redis
+
+  setup:
+    image: python:3.9.6-alpine
+    volumes:
+      - ../:/usr/src/myapp
+    working_dir: /usr/src/myapp/docker
+    command:  sh -c "python generate-nginx-conf.py"
```
john0312, line 22: Optional: Not sure if we'll have a race condition here. I think we could consider writing a Dockerfile that generates the nginx.conf.

john0312 (approved):
Both comments are optional. You can consider whether to fix them or merge first then make them a TODO and leave them as is for now.

docker/nginx-template.conf
```diff
@@ -18,20 +18,17 @@ http {{
 
     server {{
         listen 80;
```
john0312, line 20: Optional: Add `default_server` if we only have one?

PR #110, opened 2021-08-14
Add permanent storage support. Support auto-registration

wxrdnx:
* Support permanent storage of item data (by using this.helper.storeData and this.helper.loadData). Also, replace `this.items`, `this.dropItemCell` and `this.dropItemInfo` with plain objects.
* Add 'droppable' field. This field checks whether an item is droppable (For instance, furniture is not droppable, while tickers are)
* Support auto-registration instead of failing when it's not registered.
* Finish the calculation of the x and y coordinates of a dropped item.
* Fix some bugs and typos


extensions/items/standalone.mjs
```diff
@@ -261,28 +281,72 @@ class Standalone {
...
+        }
+        droppedItemCoord.y = mapCoord.y;
+        break;
     }
+    this.droppedItemCell[this.droppedItemIndex] = { "x": droppedItemCoord.x, "y": droppedItemCoord.y, "w": 1, "h": 1 };
+    this.droppedItemInfo[this.droppedItemIndex] = itemName;
+    this.droppedItemIndex++;
+
+    /* Store data into file */
+    this.helper.storeData(this._pack_stored_data());
```
john0312, line 346: `this._pack_stored_data()` is async, do you want to `await` or just make it not async?

extensions/items/standalone.mjs
```diff
@@ -261,28 +281,72 @@ class Standalone {
...
+    this.items[playerID, itemObj];
 
     /* update cell set */
-    /* TODO: calculate dropped coordinate. Currently, the dropped coordinate is the same as the player's map coordinate */
-    this.droppedItemCell.set(this.droppedItemIndex, { "x": mapCoord.x, "y": mapCoord.y, "w": 1, "h": 1 });
-    this.droppedItemInfo.set(this.droppedItemIndex, itemName);
-    this.droppedItemIndex++;
-    for (const mapName of this.helper.gameMap.getOriginalCellSetStartWith('').keys()) { // Get all available maps.
-      this.gameMap.updateDynamicCellSet(mapName, 'droppedItem' + itemName.toUpperCase(), Array.from(this.droppedItemCell.values()));
+    const mapSize = this.gameMap.getMapSize(mapCoord.mapName);
```
john0312, line 301: This is assuming all coordinates are "droppable", not sure if we can assume that.

extensions/items/standalone.mjs
```diff
@@ -293,29 +357,33 @@ class Standalone {
...
-    this.droppedItemInfo.delete(droppedItemIndex);
+    delete this.droppedItemCell[droppedItemIndex];
+    delete this.droppedItemInfo[droppedItemIndex];
 
-    /* update cell set */
-    for (const mapName of this.helper.gameMap.getOriginalCellSetStartWith('').keys()) { // Get all available maps.
-      this.gameMap.updateDynamicCellSet(mapName, 'droppedItem' + itemName.toUpperCase(), Array.from(this.droppedItemCell.values()));
-    }
+    /* Store data into file */
+    this.helper.storeData(this._pack_stored_data());
```
john0312, line 383: Same here.

extensions/items/standalone.mjs
```diff
@@ -141,11 +156,10 @@ class Standalone {
    */
   async c2s_getAllItems(playerID) {
```
john0312, line 157: Forgot to mention, the first argument for c2s API call is actually a `Player` object. You'll need to do `player.playerID` or `player.getPlayerID()` to get the playerID.

extensions/items/standalone.mjs
```diff
@@ -6,6 +6,8 @@ import fs from 'fs';
 import path from 'path';
 import CellSet from '../../common/maplib/cellset.mjs';
 
+import CellSet from '../../common/maplib/cellset.mjs';
```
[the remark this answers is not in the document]
wxrdnx: Sorry I didn't notice this mistake. Fixed!

extensions/items/standalone.mjs
wxrdnx: The function should't be an async function. Sorry for the mistake. It's fixed now.

extensions/items/standalone.mjs
wxrdnx: Same as above

wxrdnx:
The first argument for several c2s APIs are changed from playerID, which is a string, to player, which is a Player object. Please merge the PR, thanks!

wxrdnx:
Eslint checks have passed. Errors and some warnings (some fixable typos) are fixed according to eslint.

john0312 (approved):

PR #119, opened 2021-08-25
Interactive object & NPC extension

ktpss95112:
work in progress

### DIFF
--- a/common/interactive-object/client.mjs
+++ b/common/interactive-object/client.mjs
@@ -1,20 +1,29 @@
 // Copyright 2021 HITCON Online Contributors
 // SPDX-License-Identifier: BSD-2-Clause
 
+/**
+ * TODO: jsdoc
+ */
 class InteractiveObjectClientBaseClass {
   /**
    * Create the client side of the extension.
    * @constructor
    * @param {ClientExtensionHelper} helper - An extension helper object for
    * servicing various functionalities of the extension.
+   * @param {MapCoord} initialPosition - TODO
    * @param {Object} displayConfig - TODO
    * @param {Function} mapClickCallback - The function to be called when the user clicks on this interactive object.
    */
-  constructor(helper, displayConfig, mapClickCallback) {
+  constructor(helper, initialPosition, displayConfig, mapClickCallback) {
     this.helper = helper;
+    this.mapCoord = initialPosition;
     this.displayConfig = displayConfig;
 
-    // TODO: MapRenderer.registerCustomizedLayerToDraw()
+    for (const conf of displayConfig) {
+      const {zIndex, layerName, renderFunction, renderArgs} = conf;
+      this.helper.mapRenderer.registerCustomizedLayerToDraw(zIndex, layerName, renderFunction, renderArgs);
+    }
+
     // TODO: register callback on map clicking
   }
 }
--- a/common/interactive-object/client.mjs
+++ b/common/interactive-object/client.mjs
@@ -1,0 +1,22 @@
+// Copyright 2021 HITCON Online Contributors
+// SPDX-License-Identifier: BSD-2-Clause
+
+class InteractiveObjectClientBaseClass {
+  /**
+   * Create the client side of the extension.
+   * @constructor
+   * @param {ClientExtensionHelper} helper - An extension helper object for
+   * servicing various functionalities of the extension.
+   * @param {Object} displayConfig - TODO
+   * @param {Function} mapClickCallback - The function to be called when the user clicks on this interactive object.
+   */
+  constructor(helper, displayConfig, mapClickCallback) {
+    this.helper = helper;
+    this.displayConfig = displayConfig;
+
+    // TODO: MapRenderer.registerCustomizedLayerToDraw()
+    // TODO: register callback on map clicking
+  }
+}
+
+export default InteractiveObjectClientBaseClass;
--- a/common/interactive-object/client.mjs
+++ b/common/interactive-object/client.mjs
@@ -12,9 +12,10 @@ class InteractiveObjectClientBaseClass {
    * servicing various functionalities of the extension.
    * @param {MapCoord} initialPosition - TODO
    * @param {Object} displayConfig - TODO
-   * @param {Function} mapClickCallback - The function to be called when the user clicks on this interactive object.
+   * @param {Function} interactFunction - The function to be called when the user tries to
+   * interact with this object (on click, keydown, signal, etc.)
    */
-  constructor(helper, initialPosition, displayConfig, mapClickCallback) {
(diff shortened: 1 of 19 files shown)


common/interactive-object/server.mjs
```diff
@@ -0,0 +1,262 @@
...
+    const fsm = this.config.FSM; // alias
+
+    // initialize if not exist
+    if (!this.dataStore.has(playerID)) this.dataStore.set(playerID, fsm.initialState);
+
+    this._fsmWalkExit = false;
+    while (!this._fsmWalkExit) {
+      const currState = fsm.states[this.dataStore.get(playerID)];
+      if (typeof currState === 'undefined') {
+        this.dataStore.set(playerID, fsm.initialState);
```
john0312, line 112: Optional: Consider doing additional initialization and add a warning here. This is so that we can differentiate between players who are not yet initialized and when there's a misconfiguration in fsm.

common/interactive-object/server.mjs
```diff
@@ -0,0 +1,262 @@
...
+
+      const kwargs = currState.kwargs;
+      const func = 'sf_' + currState.func; // 'sf_' stands for state function
+      if (!(func in this)) {
+        console.error(`'${func}' is not a valid state function`);
+        this.dataStore.set(playerID, fsm.initialState);
+        break;
+      }
+      const nextState = await this[func](playerID, kwargs);
+      this.dataStore.set(playerID, nextState);
```
john0312, line 124: Optional: Consider adding a check here on whether player is online. If they're disconnected, then we can break.

common/interactive-object/server.mjs
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  /**
+   * Just a placeholder function to provide `exit` function in configuration.
+   * @param {String} playerID
+   * @param {Object} kwargs - TODO
+   * @return {String} - the next state
+   */
+  async sf_exit(playerID, kwargs) {
+    this._fsmWalkExit = true;
```
john0312, line 223: If I understand correctly, this class is shared by all players, so this might interfere with other players. I would recommend returning `_fsmWalkExit` with the state.

extensions/dialog/common/client.mjs
```diff
@@ -95,8 +96,8 @@ class DialogModal extends Modal {
       return btnList;
     });
 
-    if (result === false) return {cancelled: true};
-    return {token: result.dataset.token};
+    if (result) return result.dataset.token;
```
john0312, line 99: The primary reason for returning an object is so that we can differentiate between various reasons why the no choice was picked. One possible reason is because another Modal is already open, or the client disconnected, or if some error is encountered during RPC (i.e. timeout).

Currently the s2c rpc system returns an object with `error` key when something happens. A better way might be to have a system that passes exception across the RPC, but that'll take time to implement.

I'm OK with the current code change if you're sure there'll be no need to distinguish between various failure types.

common/interactive-object/client.mjs
```diff
@@ -0,0 +1,45 @@
...
+    this.displayConfig = displayConfig;
+
+    for (const conf of displayConfig) {
+      const {zIndex, layerName, renderFunction, renderArgs} = conf;
+      this.helper.mapRenderer.registerCustomizedLayerToDraw(zIndex, layerName, renderFunction, renderArgs);
+    }
+
+    // interact on click
+    this.helper.inputManager.registerCanvasOnClickMapCoord((clickedMapCoord) => {
+      // determine whether this object is clicked
```
john0312, line 30: Optional: One extra callback per NPC might slow down the client when we've a lot of NPCs. Consider having a `InteractiveObjectManager` that registers one single callback and lookup the callback through a Map. Similar optimization might be available for `registerCustomizedLayerToDraw` as well.

common/interactive-object/server.mjs
```diff
@@ -0,0 +1,262 @@
...
+        return;
+      }
+    }
+
+    if (!initialPositionValidation(this.config.initialPosition)) return;
+    if (!fsmValidation(this.config.FSM)) return;
+    if (!displayValidation(this.config.display)) return;
+
+    // TODO: store the states in database (or in `/run/small_data`)
+    this.dataStore = new Map(); // key: playerName, value: currentState
```
john0312, line 95: Note: If you're putting this in `/run/small_data`, you'll need to convert this to/from JSON object.

extensions/npc/standalone.mjs
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * TODO
+   * @param {Object} player - TODO
+   * @param {String} npcName - TODO
+   */
+  async c2s_startInteraction(player, npcName) {
+    const npc = this.NPCs.get(npcName);
+    if (typeof npc === 'undefined') return;
```
john0312, line 80: Optional: Return failure and/or log an error so it doesn't fail silently? Ditto above.

run/npc/example-npc1.json
```diff
@@ -0,0 +1,49 @@
+{
+	"enabled": true,
+	"initialPosition": {"mapName": "world1", "x": 5, "y": 15},
+  "display": [
```
john0312, line 4: Weird indent here.

common/interactive-object/server.mjs
ktpss95112: I don't see the need for online checks. If a player disconnects, the `this[func](playerID, kwargs)` will handle it (either detect a timeout or something else) by calling `this.sf_exit()`.

common/interactive-object/server.mjs
ktpss95112: Yes, exactly. Big thanks for finding this bug!
The reason for not returning `_fsmWalkExit` is to make the signature of `sf_` functions consistent. I hope their return values are always a state so that programmers don't need to find out what each `sf_` function returns.

extensions/dialog/common/client.mjs
ktpss95112: Well, I'm not sure that there will be no need to distinguish between various failure types.
I will revert this modification.

common/interactive-object/client.mjs
ktpss95112: I don't get the point of this optimization, maybe we can discuss it in the meeting.

common/interactive-object/server.mjs
ktpss95112: I'm not going to store it on disk or database. That may require a lot of effort.
Currently I store this in memory, so I use `new Map()` instead of `class DataStore`.

run/npc/example-npc1.json
ktpss95112: There are strange `'\t'`s here. Removed.

john0312 (approved):

PR #122, opened 2021-08-25
Extension playerlist basic feature done

zeze-zeze:
The way getting player list is to request server once a second, which can be more efficient if we can find a way to update player list.

### DIFF
--- a/extensions/playerlist/common/client.mjs
+++ b/extensions/playerlist/common/client.mjs
@@ -48,6 +48,7 @@ class Client {
    * @getPlayerList
    */
   async getPlayerList() {
+    // TODO(zeze-zeze): Optimize the way to update playerlist
     setInterval(() => {
       this.helper.callC2sAPI('playerlist', 'getPlayerList', 1000);
     }, 1000);
@@ -75,17 +76,6 @@ class Client {
    */
   HTMLEncode(str) {
     return $('<div/>').text(str).html();
-  }
-
-  /**
-   * Get a Broadcast message
-   * @onExtensionBroadcast
-   * @param {object} arg - Information of the broadcast message
-  */
-  onExtensionBroadcast(arg) {
-    arg.msg_from = $('<div/>').text(arg.msg_from).html();
-    arg.msg = $('<div/>').text(arg.msg).html();
-    document.getElementById('message_history').innerHTML += '<span>' + arg.msg_from + ': ' + arg.msg + '</span><br>';
   }
 }
 
--- a/extensions/playerlist/common/client.mjs
+++ b/extensions/playerlist/common/client.mjs
@@ -52,7 +52,11 @@ class Client {
         if (typeof playerID.displayName !== 'string' || !playerID.displayName.includes(document.getElementById("searchPlayer").value)) {
           return;
         }
+        // TODO(zeze-zeze): Keep a record of dom elements
+        // TODO(zeze-zeze): Move it to a separate function
         document.getElementById('playerlist').innerHTML += '<span>' + this.HTMLEncode(playerID.displayName); + '</span>';
+
+        // TODO(zeze-zeze): Use onchange callback to update playerlist instead of waiting 1 second
      });
     }, 1000);
   }
--- a/extensions/playerlist/common/client.mjs
+++ b/extensions/playerlist/common/client.mjs
@@ -40,33 +40,24 @@ class Client {
   async gameStart() {
     this.overlay = new PlayerlistOverlay(this.helper.mainUI);
     this.overlay.show(OverlayPosition.LEFT_TOP);
-    this.getPlayerList();
+    this.showPlayerList();
   }
 
   /**
-   * Get playerlist from server.
-   * @getPlayerList
+   * Show player list on left top.
+   * @showPlayerList
    */
-  async getPlayerList() {
+  showPlayerList() {
     // TODO(zeze-zeze): Optimize the way to update playerlist
     setInterval(() => {
-      this.helper.callC2sAPI('playerlist', 'getPlayerList', 1000);
+      document.getElementById('playerlist').innerHTML = '';
+      this.helper.gameState.getPlayers().forEach(playerID => {
+        if (typeof playerID.displayName !== 'string' || !playerID.displayName.includes(document.getElementById("searchPlayer").value)) {
+          return;
+        }
+        document.getElementById('playerlist').innerHTML += '<span>' + this.HTMLEncode(playerID.displayName); + '</span>';
+     });
     }, 1000);
-  }
-  
-  /**
-   * Show player list on left top.
-   * @s2c_showPlayerList
-   * @param {object} playerIDs - an array of all player ids online.
-   */
-  s2c_showPlayerList(playerIDs) {
-    document.getElementById('playerlist').innerHTML = '';
-    playerIDs.forEach(playerID => {
-      if (typeof playerID !== 'string') {
-        return;
-      }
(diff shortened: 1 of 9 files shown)


extensions/playerlist/common/client.mjs
```diff
@@ -0,0 +1,92 @@
...
+    this.overlay.show(OverlayPosition.LEFT_TOP);
+    this.getPlayerList();
+  }
+
+  /**
+   * Get playerlist from server.
+   * @getPlayerList
+   */
+  async getPlayerList() {
+    setInterval(() => {
```
john0312, line 51: Will probably add a TODO on optimizing this, as we've discussed on the extra load of polling.

extensions/playerlist/common/client.mjs
```diff
@@ -0,0 +1,92 @@
...
+    });
+  }
+
+  /**
+   * Encode the html in case of XSS
+   * @HTMLEncode
+   * @param {string} str - the string being encoded
+   */
+  HTMLEncode(str) {
+    return $('<div/>').text(str).html();
```
john0312, line 77: I'm thinking if we should unify the way we deal with XSS in this project. Will discuss in today's meeting.

extensions/playerlist/standalone.mjs
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  /**
+   * Send player list from gameState to client side.
+   * @c2s_getPlayerList
+   * @param {Player} player - player information
+   */
+  async c2s_getPlayerList(player) {
+    const playerIDs = [...this.helper.gameState.players.keys()];
+    await this.helper.callS2cAPI(player.playerID, 'playerlist', 'showPlayerList', 1000, playerIDs);
```
john0312, line 33: Have you considered simply returning the playerIDs? c2s API can return things to the client side.

extensions/playerlist/common/client.mjs
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  /**
+   * Get a Broadcast message
+   * @onExtensionBroadcast
+   * @param {object} arg - Information of the broadcast message
+  */
+  onExtensionBroadcast(arg) {
```
john0312, line 85: I don't quite get what this function is doing, can you explain?

john0312:
Please remember to run yarn install.

extensions/playerlist/common/client.mjs
zeze-zeze: I directly copied it from extension "chat", and I forgot to delete it. Thanks for reminding me.

extensions/playerlist/common/client.mjs
```diff
@@ -0,0 +1,82 @@
...
+  constructor(mainUI) {
+    const dom = document.getElementById(PLAYERLIST_DIV);
+    super(mainUI, dom);
+  }
+
+  // TODO(zeze-zeze): Expand on this class (jsdoc and implementation) and
+  // implement onPre/PostShow/Hide() as needed.
+
+  hasToolbarButton() {
+    this.toolbarButtonSrc = '/static/extensions/playerlist/playerlist.svg';
```
john0312, line 19: I think Alan changed the way toolbar button works, but not sure if your PR is getting merged first or his is getting merged first.

extensions/playerlist/common/client.mjs
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  /**
+   * Get playerlist from server.
+   * @getPlayerList
+   */
+  async getPlayerList() {
+    // TODO(zeze-zeze): Optimize the way to update playerlist
+    setInterval(() => {
+      this.helper.callC2sAPI('playerlist', 'getPlayerList', 1000);
```
john0312, line 53: As previously mentioned, it would be recommended to get the info from client side.

extensions/playerlist/standalone.mjs
```diff
@@ -0,0 +1,46 @@
...
+  async initialize() {
+  }
+
+  /**
+   * Send player list from gameState to client side.
+   * @c2s_getPlayerList
+   * @param {Player} player - player information
+   */
+  async c2s_getPlayerList(player) {
+    const playerIDs = [...this.helper.gameState.players.keys()];
```
john0312, line 32: Optional: Add a note that this function is not being used?

extensions/playerlist/common/client.mjs
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  /**
+   * Show player list on left top.
+   * @showPlayerList
+   */
+  showPlayerList() {
+    // TODO(zeze-zeze): Optimize the way to update playerlist
+    setInterval(() => {
+      document.getElementById('playerlist').innerHTML = '';
```
john0312, line 53: Optional: Consider making this a separate function, so that you can call this in onchange event of the search bar.

extensions/playerlist/common/client.mjs
```diff
@@ -0,0 +1,73 @@
...
+      this.helper.gameState.getPlayers().forEach(playerID => {
+        if (typeof playerID.displayName !== 'string' || !playerID.displayName.includes(document.getElementById("searchPlayer").value)) {
+          return;
+        }
+        document.getElementById('playerlist').innerHTML += '<span>' + this.HTMLEncode(playerID.displayName); + '</span>';
```
john0312, line 58: Optional: Consider keeping a record of what dom elements already exist in the player list, then add/delete as required. This will prevent us from adding/removing every `<span>` element every second, reducing the load for better UI.

extensions/playerlist/in-div.ejs
```diff
@@ -0,0 +1,13 @@
+<style type="text/css">
+  #playerlist span {
+    margin: 5px;
+  }
+</style>
+
+<div id="playerlist-overlay">
+  <h2>Player List</h2>
+  <input type="text" id="searchPlayer">
```
john0312, line 9: Optional: Consider adding an onchange event callback so users will not have to wait 1 second for the update.

extensions/playerlist/common/client.mjs
```diff
@@ -0,0 +1,73 @@
...
+     });
+    }, 1000);
```
john0312, line 60: Optional: You can register player info change callback with `GameState`, so we can prevent update when there's no change.

john0312 (approved):
Some optional comments, you can merge first then consider if you want to implement the recommendations.

extensions/playerlist/common/client.mjs
john0312: I think he just merged, you might have to update.

extensions/playerlist/common/client.mjs
zeze-zeze: I think it is not more efficient to register player info change because we have to call the callback whenever anybody moves.

extensions/playerlist/in-div.ejs
zeze-zeze: Added to TODO

extensions/playerlist/common/client.mjs
zeze-zeze: Added to TODO

extensions/playerlist/common/client.mjs
zeze-zeze: Added to TODO

extensions/playerlist/standalone.mjs
zeze-zeze: Deleted

john0312 (approved):

PR #133, opened 2021-09-05
more chat extension features

zeze-zeze:
1. interaction between playerlist and chat extension
2. nearby message
3. some changes to the chat layout and coding style


extensions/playerlist/in-div.ejs
```diff
@@ -4,6 +4,13 @@
   }
 </style>
 
+<script type="text/javascript">
+  function startPrivateMessage(playerID) {
+    console.log(playerID);
+    document.getElementById("message_to").value = playerID;
```
john0312, line 10: Optional: Would recommend moving this to client.mjs to make it simpler when the apply the UI changes.

john0312 (approved):

PR #134, opened 2021-09-11
Point system integration

s3131212:
1. Implement the basic point system integration.
2. Start the client side of the extension (i.e. `gameStart`) after the socket.io authentication is completed.

I have not yet implemented the UI for transferring points, since I have no idea how it should look like :( The function is completed, though.


extensions/point-system/standalone.mjs
```diff
@@ -0,0 +1,65 @@
...
+    return {inDiv: 'in-div.ejs'};
+  }
+
+  /**
+   * Notify other users to update the points.
+   * @param {Player} player - player information
+   * @param {string} uid - The user to be notified.
+   */
+  async c2s_notifyUpdatePoints(player, uid) {
+    return await this.helper.callS2cAPI(uid, 'point-system', 'updatePoints', 5000);
```
john0312, line 45: Handle the case when target user is not online or use a try to ignore any exception?

john0312 (approved):

PR #135, opened 2021-09-14
Add player context menu

s3131212:
As title.

One can use the following code to register a context menu item in an extension.
```js
// This item is only shown when right-clicking on other players.
this.helper.mainUI.contextMenu.addToOthersMenu('test', (player) => console.log(player));

// This item is only shown when right-clicking on oneself.
this.helper.mainUI.contextMenu.addToSelfMenu('test', (player) => console.log(player));
```

I prefer merging this PR before PR #134, so that I can implement the point system in a single PR.

### DIFF
--- a/sites/game-client/client.ejs
+++ b/sites/game-client/client.ejs
@@ -35,7 +35,7 @@ absolute URL.
             <p id="error-message"></p>
           </div>
           <p id="error-handler-message"></p>
-        </div> 
+        </div>
         <% inDiv.forEach((i) => { %>
           <%- include(i); %>
         <% }); %>
@@ -58,6 +58,13 @@ absolute URL.
       <div id="modal-background" class="modal-background">
       </div>
       <div id="modal-container" class="modal-container">
+      </div>
+      <!-- context menu -->
+      <div id="context-menu-other" class="context-menu">
+        <ul id="context-menu-list-other" class="context-menu-list"></ul>
+      </div>
+      <div id="context-menu-self" class="context-menu">
+        <ul id="context-menu-list-self" class="context-menu-list"></ul>
       </div>
     </div>
 
@@ -149,6 +156,8 @@ absolute URL.
             game.errorModal.displayError(reason, "Please refresh to reconnect.");
           }
         });
+
+        game.mainUI.createContextMenu(game.gameState, game.mapRenderer);
 
         function onGameTick () {
           game.mapRenderer.draw();
--- a/sites/game-client/ui/context-menu.mjs
+++ b/sites/game-client/ui/context-menu.mjs
@@ -26,7 +26,7 @@ class ContextMenu {
 
     // bind the contextmenu and click event.
     $('#mapcanvas').on('contextmenu', this.canvasOnContextMenu.bind(this));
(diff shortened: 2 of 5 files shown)


sites/game-client/ui/context-menu.mjs
```diff
@@ -0,0 +1,157 @@
...
+        } else if (this.focusedPlayer.playerID !== this.playerInfo.playerID && this.othersMenu.size > 0) {
+            div = CONTEXT_MENU_OTHER_DIV;
+        } else {
+            this.focusedPlayer = null;
+            return;
+        }
+
+        let menu = document.getElementById(div);
+        if (menu.style.display == "block") {
+            this.hideMenu();
```
john0312, line 63: Maybe a comment on what this block is for? i.e. menu was already open.

john0312: Also, if the menu is already open, maybe we should close it first then open the current one?

sites/game-client/ui/context-menu.mjs
```diff
@@ -0,0 +1,157 @@
...
+        // Stop if the menu is empty.
+        let div;
+        if (this.focusedPlayer.playerID === this.playerInfo.playerID && this.selfMenu.size > 0) {
+            div = CONTEXT_MENU_SELF_DIV;
+        } else if (this.focusedPlayer.playerID !== this.playerInfo.playerID && this.othersMenu.size > 0) {
+            div = CONTEXT_MENU_OTHER_DIV;
+        } else {
+            this.focusedPlayer = null;
+            return;
+        }
```
john0312, line 59: Optional: I would generally recommend splitting the logic between opening the context menu and triggering it from events from the canvas. The reason is because there may be multiple causes for triggering the context menu, such as people right clicking on the username in the chat overlay. This change can be delayed to later when we add the trigger for context menu in chat overlay.

sites/game-client/client.ejs
```diff
@@ -59,6 +59,13 @@ absolute URL.
       </div>
       <div id="modal-container" class="modal-container">
       </div>
+      <!-- context menu -->
```
john0312, line 62: Optional: We can also consider using off the shelf library for the actual context menu.

john0312 (approved):

john0312:
You can merge this PR first and leave the comments as TODOs.

sites/game-client/client.ejs
s3131212: I believe that since the context menu has to be deeply integrated into our current design, it's quite difficult to implement it using off-the-shelf libraries. 
Given that the current solution works, I prefer to keep it as it currently is. However, if we later find it difficult to maintain the code, I'm open to using libraries.

sites/game-client/ui/context-menu.mjs
s3131212: I'm not sure what's the common design of context menu, but at least on my machine, the second right-click would close the existing context menu.

s3131212:
I added a function to support opening the context menu from place other than the canvas.
Use `data-player-context-menu=player id` to trigger the context menu, as shown in the playerlist extension.

sites/game-client/client.ejs
john0312: Ok, sure, let's leave the custom implementation as is.

PR #139, opened 2021-09-18
refactor gateway-service, add player smooth movement, refactor input manager

ktpss95112:
Still working in progress
9/24 updated: done

### DIFF
--- a/common/gamelib/move-check.mjs
+++ b/common/gamelib/move-check.mjs
@@ -7,31 +7,31 @@ const PLAYER_MOVE_DISTANCE_LIMIT = 1; // cell
 /**
  * TODO
  * @param {Player} oldPlayerData - TODO
- * @param {PlayerSyncMessage} newPlayerData - TODO
+ * @param {PlayerSyncMessage} updateMessage - TODO
  * @param {GameMap} gameMap - TODO
  * @return {Boolean}
  */
-function checkPlayerMove(oldPlayerData, newPlayerData, gameMap) {
+function checkPlayerMove(oldPlayerData, updateMessage, gameMap) {
   let canMove = true;
-  canMove = canMove && _speedCheck(oldPlayerData, newPlayerData);
-  canMove = canMove && _borderAndWallCheck(oldPlayerData, newPlayerData, gameMap);
+  canMove = canMove && _speedCheck(oldPlayerData, updateMessage);
+  canMove = canMove && _borderAndWallCheck(oldPlayerData, updateMessage, gameMap);
   return canMove;
 }
 
 /**
  * TODO
  * @param {Player} oldPlayerData - TODO
- * @param {PlayerSyncMessage} newPlayerData - TODO
+ * @param {PlayerSyncMessage} updateMessage - TODO
  * @return {Boolean}
  */
-function _speedCheck(oldPlayerData, newPlayerData) {
+function _speedCheck(oldPlayerData, updateMessage) {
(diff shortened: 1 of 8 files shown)


common/gamelib/move-check.mjs
```diff
@@ -0,0 +1,68 @@
...
+const PLAYER_MOVE_DISTANCE_LIMIT = 1; // cell
+
+/**
+ * TODO
+ * @param {Player} oldPlayerData - TODO
+ * @param {PlayerSyncMessage} newPlayerData - TODO
+ * @param {GameMap} gameMap - TODO
+ * @return {Boolean}
+ */
+function checkPlayerMove(oldPlayerData, newPlayerData, gameMap) {
```
john0312, line 14: It seems that `newPlayerData` is `PlayerSyncMessage` type while `oldPlayerData` is `Player` type. Maybe we can rename the variable for better readability?

common/gamelib/player.mjs
```diff
@@ -57,11 +85,15 @@ class Player {
...
-    if (msg.mapCoord !== undefined) this.mapCoord = msg.mapCoord;
+    if (msg.mapCoord !== undefined) {
+      this._previousMapCoord = (this.mapCoord ?? msg.mapCoord); // for smoothly move the player
+      this.mapCoord = msg.mapCoord;
+    }
     if (msg.facing !== undefined) this.facing = msg.facing;
     if (msg.displayName !== undefined) this.displayName = msg.displayName;
     if (msg.displayChar !== undefined) this.displayChar = msg.displayChar;
     if (msg.removed !== undefined) this.removed = msg.removed;
+    this.lastMovingTime = Date.now();
```
john0312, line 96: Maybe set `this.lastMovingTime` only if `this.mapCoord` changed? Considering `updateFromMessage()` is not limited to only moving the user.

services/gateway/gateway-service.mjs
```diff
@@ -235,57 +236,28 @@ class GatewayService {
...
-    if (socket.playerID !== origMsg.playerID) {
-      console.error(`Player '${origMsg.playerID}' tries to update player '${socket.playerID}'s data, which is invalid.`);
+  async onPlayerUpdate(socket, updateMsg) {
+    if (socket.playerID !== updateMsg.playerID) {
+      console.error(`Player '${updateMsg.playerID}' tries to update player '${socket.playerID}'s data, which is invalid.`);
       return;
     }
 
-    const msg = PlayerSyncMessage.fromObject(socket.playerData);
-    msg.facing = origMsg.facing;
```
john0312, line 248: I think it makes sense to explicitly copy over the fields that we want to allow the users to update (the way it's done in the deleted code). This way, whenever a new field is added, there's lesser chance for us to introduce logical bug that allows user to manipulate the field without check.

john0312 (approved):

services/gateway/gateway-service.mjs
ktpss95112: One of the purposes to use `PlayerSyncMessage` class is to restrict the fields that can be updated. I think it is unnecessary for the server to construct the update message again.

common/gamelib/player.mjs
ktpss95112: fixed

common/gamelib/move-check.mjs
ktpss95112: fixed

PR #140, opened 2021-09-19
feat(ext-iframe): support overlay iframe with map configuration

canyugs:
open/close iframe by map configuration

### DIFF
--- a/extensions/iframe/common/client.mjs
+++ b/extensions/iframe/common/client.mjs
@@ -1,0 +1,61 @@
+// Copyright 2021 HITCON Online Contributors
+// SPDX-License-Identifier: BSD-2-Clause
+
+import Overlay from '/static/sites/game-client/ui/overlay.mjs';
+import OverlayPosition from '/static/sites/game-client/ui/overlay-position.mjs';
+
+class IframeOverlay extends Overlay {
+  constructor(mainUI) {
+    const dom = document.getElementById('iframe-overlay');
+    super(mainUI, dom);
+    this.iframe = dom.querySelector('iframe');
+    this.hide();
+    this.lastSrc = null;
+  }
+
+  updateIframe(src) {
+    if (this.lastSrc === src) return;
+    if (src === null) {
+      this.hide();
+      this.lastSrc = null
+      return;
+    }
+
+    if (!this.isOpen) {
+      this.iframe.src = src;
+      this.lastSrc = src;
+      this.show(OverlayPosition.LEFT_BOTTOM);
+      return;
+    }
+  }
+}
+
+/**
+ * This class is the browser/client side of an extension.
+ * One instance is created for each connected player.
+ */
+class Client {
+  /**
+   * Create the client side of the extension.
+   * @constructor
+   * @param {ClientExtensionHelper} helper - An extension helper object for
+   * servicing various functionalities of the extension.
+   */
+  constructor(helper) {
+    this.helper = helper;
+  }
+
+  async gameStart() {
+    this.overlay = new IframeOverlay(this.helper.mainUI);
+  }
+
+  onSelfPlayerUpdate(msg) {
+    const map = this.helper.getMap();
+    if (typeof map === 'object') {
+     let src = map.getCell(msg.mapCoord, 'videoIframe');
+     this.overlay.updateIframe(src);
+    }
+  }
+}
+
+export default Client;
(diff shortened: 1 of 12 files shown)


extensions/iframe/common/client.mjs
```diff
@@ -0,0 +1,73 @@
...
+    this.isOpen = true;
+  }
+
+  close() {
+    this.hide();
+    this.isOpen = false;
+  }
+}
+
+class IframeModal extends Modal {
```
john0312, line 31: Not sure if need IframeModal. If we indeed use IframeModal, then the player can't move when the IframeModal shows up. Currently we don't have a use case for this, but it doesn't hurt to leave it there with a comment on it.

extensions/iframe/common/client.mjs
```diff
@@ -0,0 +1,73 @@
...
+  display(src, attributes = {}) {
+    if (this.isOpen) return;
+    this.iframe.src = src;
+    this.show(OverlayPosition.LEFT_BOTTOM);
+    this.isOpen = true;
+  }
+
+  close() {
+    this.hide();
+    this.isOpen = false;
```
john0312, line 27: Maybe we can implement onPostHide() so that if someone else calls hide() then isOpen will be correct as well.

extensions/iframe/common/client.mjs
```diff
@@ -0,0 +1,73 @@
...
+    this.iframe = dom.querySelector('iframe');
+    this.hide();
+    this.isOpen = false;
+  }
+
+  display(src, attributes = {}) {
+    if (this.isOpen) return;
+    this.iframe.src = src;
+    this.show(OverlayPosition.LEFT_BOTTOM);
+    this.isOpen = true;
```
john0312, line 22: Maybe implement onPostShow() so that even if someone else called show(), isOpen will be correct.

extensions/iframe/common/client.mjs
```diff
@@ -0,0 +1,73 @@
...
+    this.overlay = new IframeOverlay(this.helper.mainUI);
+    this.overlay.close();
+  }
+
+  onSelfPlayerUpdate(msg) {
+    const map = this.helper.getMap();
+    const iframeCellSet = map.getOriginalCellSet('iframeLocationFixed').get('world1');
+    const {mapName, x, y} = msg.mapCoord;
+    const currentMap = new MapCoord(mapName, x, y);
+    if (currentMap.insideCellSet(iframeCellSet)) {
```
john0312, line 65: As discussed, maybe use map.getCell()?

package.json
```diff
@@ -29,6 +29,7 @@
     "typescript": "^4.2.4"
   },
   "scripts": {
-    "postinstall": "cp ./git-hook .git/hooks/pre-commit"
+    "postinstall": "cp ./git-hook .git/hooks/pre-commit",
+    "redis": "docker run -d --rm --name dev-redis -p 6379:6379 redis"
```
john0312, line 33: We might not need this?

extensions/iframe/common/client.mjs
canyugs: got it!

package.json
canyugs: my mistake, only used by myself,

extensions/iframe/common/client.mjs
canyugs: I think there will be a close button when IframModal is active. I decide to remove it  and wait for the real cases.

extensions/iframe/common/client.mjs
canyugs: Try to get this done....

extensions/iframe/common/client.mjs
canyugs: Try to get this done....

sites/game-client/ui/overlay.mjs
```diff
@@ -141,7 +145,10 @@ class Overlay {
    * This method can be implemented in the derived class if the derived class
    * wishes to handle this event.
    */
-  onPostHide() { return true; }
+  onPostHide() {
+    this.isOpen = false;
```
john0312, line 149: There is no guarantee that the derived class will call the base class' onPostHide(). To better maintain this.isOpen, it would be recommended to put the update code in show() and hide() above.

extensions/iframe/common/client.mjs
canyugs: I realized that I forget about Class can do, thx

extensions/iframe/common/client.mjs
canyugs: I realized that I forget about Class can do, thx

john0312 (approved):

PR #142, opened 2021-09-25
Fix watermark

zeze-zeze:
The function name "_drawWatermark" had been changed into "_drawWaterMark", so it couldn't be called.


john0312 (approved):
I would prefer `_drawWatermark` because it's "watermark" not "water mark". However, it's more of an English issue than a code issue, so I'm OK either way.

john0312 (approved):

PR #144, opened 2021-09-25
Command for Announcement

fanlan1210:
now announcement can triggered by command in the chat extension.

not supported yet:
* permission check
* announcement timeout (?)

### DIFF
--- a/extensions/chat/common/client.mjs
+++ b/extensions/chat/common/client.mjs
@@ -98,7 +98,7 @@ Usage: !/<Command> <arg1> <arg2> ...
       } else {
         this.helper.callC2sAPI(null, 'teleport', 5000, mapCoord);
       }
-    } else if (cmd.split(' ')[0] === '!/announce') {
+    } else if (cmd.split(' ')[0] === '!/announce') { // TODO: restrict this using permission to admin
       const msg = cmd.substring(10);
       if (msg.trim() === '') {
         document.getElementById('message_history').innerHTML += '<span>Invalid message</span><br>';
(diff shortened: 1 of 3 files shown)


extensions/chat/common/client.mjs
```diff
@@ -97,6 +98,15 @@ Usage: !/<Command> <arg1> <arg2> ...
       } else {
         this.helper.callC2sAPI(null, 'teleport', 5000, mapCoord);
       }
+    } else if (cmd.split(' ')[0] === '!/announce') {
```
john0312, line 101: Optional: Maybe leave a TODO that we'll need to restrict this to admin?

extensions/chat/common/client.mjs
```diff
@@ -150,10 +160,16 @@ Usage: !/<Command> <arg1> <arg2> ...
    * @param {object} args - Information of the broadcast message
   */
   onExtensionBroadcast(args) {
-    args.msg_from = $('<div/>').text(args.msg_from).html();
-    args.msg = $('<div/>').text(args.msg).html();
-    document.getElementById('message_history').innerHTML += '<span>' + args.msg_from + ': ' + args.msg + '</span><br>';
+    if (args.type === 'announcement') {
```
john0312, line 163: I think we'll need to implement filtering on the server side.

john0312 (approved):

extensions/chat/common/client.mjs
fanlan1210: done.

PR #148, opened 2021-09-25
Remove the default value for iframe to prevent loading the /client.html twice

s3131212:
As title.
I feel honored to have the PR with the shortest edit distance :)


john0312 (approved):
lmao, nice catch.

PR #152, opened 2021-09-26
Bulletin npc

zeze-zeze:
1. Check permission in state
2. Edit dialog

### DIFF
--- a/common/interactive-object/server.mjs
+++ b/common/interactive-object/server.mjs
@@ -259,6 +259,21 @@ class InteractiveObjectServerBaseClass {
 
     return this.sf_exit(playerID, {next: this.dataStore.get(playerID)});
   }
+
+  async sf_checkPermission(playerID, kwargs) {
+    // TODO: Give edit permission according to JWT token
+    const permission = 'editor';
+
+    const {options} = kwargs;
+    for (const [identity, nextState] of Object.entries(options)) {
+      if (permission === identity) {
+        return nextState;
+      }
+    }
+    // No identity match
+    throw 'No identity match';
+  }
+
   /**
    * Just a placeholder function to provide `exit` function in configuration.
    * @param {String} playerID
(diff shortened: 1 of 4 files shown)


john0312:
If I understand correctly, previously we wanted to go with the editable variable (instead of editable fsm) approach?

common/interactive-object/server.mjs
```diff
@@ -259,6 +262,52 @@ class InteractiveObjectServerBaseClass {
...
+
+  /**
+   * Check permission by JWT token and go to corresponding state
+   * @param {String} playerID
+   * @param {Object} kwargs - TODO
+   * @return {String} - the next state
+   */
+  async sf_checkPermission(playerID, kwargs) {
+    const permission = await this.helper.getToken(playerID);
+    permission.scope.push('online_bulletin_editor');
```
john0312, line 274: Did you leave this here for testing?

common/interactive-object/server.mjs
```diff
@@ -259,6 +262,52 @@ class InteractiveObjectServerBaseClass {
...
+    const {options} = kwargs;
+    for (const [identity, nextState] of Object.entries(options)) {
+      if (permission.scope.includes(identity)) {
+        return nextState;
+      }
+    }
+    // No identity match
+    throw 'No identity match';
```
john0312, line 282: We might need another kwargs parameter for this, so that the user can catch the conditions whereby no scope is matched.

common/interactive-object/server.mjs
zeze-zeze: Yes. Because there is no scope called "online_bulletin_editor" now, I will add a TODO to remember to delete it later.

zeze-zeze:
I declare a variable "editedFSM" for a copy of fsm, and the original "config.FSM" is not changed.

common/interactive-object/server.mjs
zeze-zeze: Now the exception is catched [here](https://github.com/canyugs/hitcon-online/blob/main/common/interactive-object/server.mjs#L142)

PR #160, opened 2021-09-28
ui(ext_jitsi): apply basic ui

canyugs:
1. Apply style to Jitsi
2. Add some examples in in-div.ejs
3. Add css order for non-active user


s3131212 (approved):
LGTM. I prefer merging this before #157 so that I can rebase onto this and continue to work on the UI.

PR #166, opened 2021-10-06
Bulletin NPC Update

zeze-zeze:
1. Separate editable dialog from FSM
2. Set default value to checkPermission


common/interactive-object/server.mjs
```diff
@@ -297,9 +300,10 @@ class InteractiveObjectServerBaseClass {
     if (Array.isArray(dialogs)) d = randomChoice(dialogs);
 
     const result = await this.helper.callS2cAPI(playerID, 'dialog', 'showDialogWithPrompt', 60*1000, this.objectName, d, buttonText);
-    console.log(result);
     if (result.msg) {
-      this.editedFSM.states[nextState].kwargs.dialogs = result.msg;
+      if (this.editableDialogs) {
+        this.editableDialogs[this.config.FSM.states[nextState].kwargs.dialogs] = result.msg;
```
john0312, line 305: Consider allowing the json file to specify the which variable name or state to fetch from. i.e.
```
let varName = '';
if (typeof kwargs.varName === 'string') varName = kwargs.varName;
else if (typeof kwargs.editState === 'string' ) varName = this.config.FSM.states[kwargs.editState].kwargs.dialogsVar;
else varName = this.config.FSM.states[nextState].kwargs.dialogsVar;
// ...
this.globalVars[varName] = result.msg;
```

extensions/npc/standalone.mjs
```diff
@@ -4,6 +4,16 @@
 import fs from 'fs';
 import InteractiveObjectServerBaseClass from '../../common/interactive-object/server.mjs';
 
+/**
+ * TODO: jsdoc
+ */
+class BulletinNPC extends InteractiveObjectServerBaseClass {
```
john0312, line 10: I would recommend making this feature available to all interactive objects, and we'll not have to deal with different NPC class here.

extensions/npc/standalone.mjs
```diff
@@ -25,9 +35,15 @@ class Standalone {
 
     // create individual NPC
     fs.readdirSync('../run/npc').forEach((file) => {
+      if (file[0] == ".") return;
       const npcName = file.slice(0, -('.json'.length));
-      const npc = new SingleNPC(helper, npcName, `../run/npc/${file}`);
-      this.NPCs.set(npcName, npc);
+      if (npcName === "BulletinNPC") {
```
john0312, line 40: We want to prevent hard coding `npcName` as much as possible, this way we'll have flexibility in handling multiple different bulletin NPCs.

extensions/npc/standalone.mjs
```diff
@@ -4,6 +4,16 @@
 import fs from 'fs';
 import InteractiveObjectServerBaseClass from '../../common/interactive-object/server.mjs';
 
+/**
+ * TODO: jsdoc
+ */
+class BulletinNPC extends InteractiveObjectServerBaseClass {
+  constructor(helper, objectName, configFilePath) {
+    super(helper, objectName, configFilePath);
+    this.editableDialogs = {"<editable1>": "Welcome to HITCON online bulletin"};
```
john0312, line 13: I would recommend naming it `globalVars` instead of `editableDialogs`. This way, this object will be a generic feature for storing any state information for the NPC, instead of being specific to bulletin dialogs, and we'll have more reasons to make it part of the standard InteractiveObject code base.

extensions/chat/common/client.mjs
```diff
@@ -159,6 +159,15 @@ Usage: !/<Command> <arg1> <arg2> ...
     this.generateMessageBox(this.HTMLEncode(args.msg_from), this.HTMLEncode(args.msg), 'nearby'));
   }
 
+  /**
+   * Get nearby message from server
+   * @s2c_getNearbyMessage
+   * @param {object} args - Information of the nearby message including msg_from and msg
+   */
+  s2c_getNearbyMessage(args) {
+    document.getElementById('message_history').innerHTML += '<span>Nearby Message From ' + this.HTMLEncode(args.msg_from) + ': ' + this.HTMLEncode(args.msg) + '</span><br>';
```
john0312, line 168: Please use `generateMessageBox` for better styling.

john0312 (approved):

PR #170, opened 2021-10-09
Fix some Jitsi bugs and add RNNoise

s3131212:
Fix:
1. The `<video>` is not shown when the webcam is unmuted.
2. The `overflow: hidden` cause the voice indicator (using `box-shadow`) failed to display.
3. Check if the previous conference was fully removed before creating a new one.

Add:
1. RNNoise noise reduction.


extensions/jitsi/common/client.mjs
```diff
@@ -125,7 +125,12 @@ class Client {
   /**
    * Start the Jitsi Meeeting
    */
-  startMeeting(meetingName, password) {
+  async startMeeting(meetingName, password) {
+    // if this.jitsiObj is set, the previous meeting has not yet ended.
+    while (this.jitsiObj) {
```
john0312, line 130: optional: Consider awaiting a Promise that is resolved when this.jitsiObj is destroyed. Note that this path can be a little bit convoluted and the benefits may not be so apparent so consider making this a TODO.

john0312 (approved):

extensions/jitsi/common/rnnoise/COPYRIGHT.md
```diff
@@ -0,0 +1,3 @@
+`RnnoiseProcessor.js` and `index.js` are reproduced from the [official Jitsi Meet client](https://github.com/jitsi/jitsi-meet/blob/master/react/features/stream-effects/rnnoise), published under [Apache License 2.0](https://github.com/jitsi/jitsi-meet/blob/master/LICENSE).
```
john0312, line 1: I would recommend using submodule and symbolic links to add these files so that we know what revisions they're from.

PR #175, opened 2021-10-11
Allow scope for test token to be specified from query parameter and fix authentication error not showing up

john0312:


s3131212 (approved):

services/auth/AuthServer.mjs
```diff
@@ -100,11 +105,12 @@ class AuthServer {
               }
               return result;
             }
+            const scopeStr = req.query.scope?req.query.scope:'';
             const payload = {};
             payload.iss = 'https://hitcon.org';
             payload.sub = genRandomStr(5);
             payload.iat = Math.floor(Date.now() / 1000);
-            payload.scope = ['point_system'];
+            payload.scope = scopeStr.split('|');
```
john0312, line 113: I think I might need to handle the case whereby scopeStr === ''

PR #176, opened 2021-10-11
Add s2s APIs to items

john0312:


extensions/items/standalone.mjs
```diff
@@ -154,11 +176,8 @@ class Standalone {
    *   }
    * }
    */
-  async c2s_getAllItems(player) {
-    /* player has not registered yet */
-    if (!(player.playerID in this.items)) {
-      this.items[player.playerID] = {};
-    }
+  async c2s_getAllDroppedItems(player) {
+    this.ensurePlayer(player.playerID);
```
canyugs: Is `_` missing ?

canyugs (approved):
Will we use tileset for Item image?

PR #183, opened 2021-10-16
Include XSS santize library and santize some extensions

zeze-zeze:


extensions/jitsi/common/client.mjs
```diff
@@ -165,23 +165,28 @@ class Client {
       this.stopMeeting();
       return;
     }
+    
+    // Sanitize to prevent XSS
+    const san_meetingName = filterXSS(meetingName);
```
john0312, line 170: I don't think we need to filter the meeting name, it's not shown anywhere.

extensions/notification/common/client.mjs
```diff
@@ -28,7 +28,8 @@ class Client {
    * Note: This function returns immediately without waiting for the notification to disappear.
    */
   async s2c_showNotification(message, timeout) {
-    this.helper.mainUI.showNotification(message, timeout);
+    const san_message = filterXSS(message);
```
john0312, line 31: Consider dealing with this directly in the mainUI.showNotification()/showAnnouncement() method, so that we don't have to change every user of it.

sites/game-client/client.ejs
```diff
@@ -18,6 +18,7 @@ absolute URL.
     <link rel="stylesheet" href="/static/sites/game-client/client.css">
     <script src="https://cdnjs.cloudflare.com/ajax/libs/socket.io/4.1.2/socket.io.min.js" integrity="sha512-2ho+gY0H62N4Z9oxD422L2ZYXiOMq9l+Aub1QNz4Z4UDOtDVxz4vk6O6R3Hqqc9Y7qSvZFl7cy3+eZ/ITnUrLA==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
     <script src="https://code.jquery.com/jquery-3.6.0.min.js" integrity="sha256-/xUj+3OJU5yExlq6GSYGSHk7tPXikynS7ogEvDej/m4=" crossorigin="anonymous"></script>
+    <script src="https://rawgit.com/leizongmin/js-xss/master/dist/xss.js"></script>
```
john0312, line 21: Consider getting it here:
https://cdnjs.com/libraries/js-xss
The script tag copied from cdnjs.com will include the integrity hash for better security.

extensions/notification/common/client.mjs
zeze-zeze: Btw, it seems that we didn't check the type of `msg` to be String before we sanitize them.

PR #207, opened 2021-10-24
Fix some bugs in the production code

s3131212:
Change:
1. Add environment variable for the `run` dir, try `RUN_DIR="../../hitcon-cat-adventure/run" docker-compose up` or `docker-compose start ... -e RUN_DIR=../../hitcon-cat-adventure/run`
2. The standalone extension services now wait for all extension to be created before initialize the extension. This fix the race condition that, sometimes the extension's `initialize` try to contact the extensions not yet loaded, or update the game state.  
3. Move the nginx config generation out of the container. One should run `generate-nginx-conf.py` to explicitly update the nginx config file.
4. Fix some minor bugs.


s3131212:
Merge conflict fixed.

common/rpc-directory/multi-process-RPC-directory.mjs
```diff
@@ -115,7 +115,7 @@ class MultiProcessRPCDirectory extends Directory {
         methodName: methodName,
         args: JSON.stringify(args)
       }, {deadline: new Date(Date.now() + 5000)});
-      return JSON.parse(ret.response);
+      return ret.response ? JSON.parse(ret.response) : null;
```
john0312, line 118: Log an error for invalid JSON? It shouldn't happen.

run-all.sh
```diff
@@ -1,6 +1,18 @@
...
+  pushd "${rundir}/config" >/dev/null
+  rundir=`pwd`
+  popd >/dev/null
+else
+  rundir="${cpath}/run/config"
+fi
+
+pushd "${path}" >/dev/null
+NODE_CONFIG_DIR="${rundir}" yarn node "${cpath}/start-all.mjs" --gateway-service gatewayServer
+popd >/dev/null
```
john0312, line 18: Optional: Please add new line.

run/config/production.json
```diff
@@ -44,19 +44,19 @@
...
-      "setting": "127.0.0.1:5026",
-      "iobj-lib": "127.0.0.1:5027",
-      "map-portal": "127.0.0.1:5028",
-      "MapWatermark": "127.0.0.1:5029",
-      "notification": "127.0.0.1:5030",
+      "dialog": "127.0.0.1:5021",
+      "setting": "127.0.0.1:5022",
+      "bombman": "127.0.0.1:5023",
+      "iobj-lib": "127.0.0.1:5024",
+      "items": "127.0.0.1:5025",
```
john0312, line 51: We might have to add 'escape-room'

john0312 (approved):

common/rpc-directory/multi-process-RPC-directory.mjs
s3131212: The `ret.response` would be empty if the (remote) function doesn't return anything, causing the `JSON.parse` to throw error.

PR #222, opened 2021-11-02
Map renderer performance

ktpss95112:


sites/game-client/map-renderer.mjs
```diff
@@ -211,6 +226,167 @@ class MapRenderer {
...
+                canvasX,
+                canvasY,
+                MAP_CELL_SIZE,
+                MAP_CELL_SIZE,
+            );
+          }
+
+          // avoid blocking the event loop
+          await new Promise((resolve) => {
+            setTimeout(resolve, 0);
```
john0312, line 327: Optional: I would recommend doing double stencil so that it won't be possible for the user to see partially drawn background.

sites/game-client/map-renderer.mjs
```diff
@@ -211,6 +226,167 @@ class MapRenderer {
...
+  async generateBackground() {
+    // There should be at most one thread doing this job in client's browser.
+    if (this._generateBackgroundLock !== undefined) {
+      return;
+    }
+    this._generateBackgroundLock = true;
+
+    // If the current map is not changed, no need to rerender.
+    const mapName = this.viewerPosition.mapName;
+    if (this._currentBackgroundMapName === mapName) {
```
john0312, line 289: Optional: Note that it's possible for cellSet to change the background layers, but it only happens if drop item or bomb man is enabled. If they're enabled, I would recommend that we have a way to mark the background layer as dirty so it can be redrawn.

john0312 (approved):

PR #223, opened 2021-11-03
Fix 駭客貓歷險記 issues in the production mode

s3131212:
1. Make sure that the terminal server works in the production mode [1].
2. Add error handlers in the FSM of 駭客貓歷險記, so that error status won't crash the team manager.
3. Fix some random bugs in 駭客貓歷險記.
4. Fix an issue in dialog, which crash when the `btns` value is not iterable (i.e. HTMLCollections).

[1] Since the terminal server requires access to docker, and it's difficult to access docker in docker container (linking docker.sock into the container is usually a terrible idea), I believe running terminal server separately, not putting it in `/docker/docker-compose.yml` is a better solution. However, this indicates that the `online` container can access the host network :( 
Calling for better ideas.

### DIFF
--- a/extensions/escape-game/standalone.mjs
+++ b/extensions/escape-game/standalone.mjs
@@ -9,6 +9,7 @@ const protoLoader = require('@grpc/proto-loader');
 const randomBytes = require('crypto').randomBytes;
 const path = require('path');
 const jwt = require('jsonwebtoken');
+const config = require('config');
 
 import fs from 'fs';
 import {promisify} from 'util';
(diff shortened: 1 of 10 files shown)


extensions/escape-game/standalone.mjs
```diff
@@ -559,7 +617,7 @@ class Team {
       try {
         let ret = await promisify(this.terminalServerGrpcService.CreateContainer.bind(this.terminalServerGrpcService))({
           imageName: defaultTerminal.imageName
-        }, {deadline: new Date(Date.now() + 5000)});
+        }, {deadline: new Date(Date.now() + 20000)});
```
john0312, line 620: Maybe 1 minute?

john0312 (approved):

PR #225, opened 2021-11-05
Code cleaning

ktpss95112:


common/gamelib/player.mjs
```diff
@@ -250,11 +250,15 @@ class PlayerSyncMessage {
...
+    if (this.facing !== undefined && !['U', 'D', 'L', 'R'].includes(this.facing)) {
+      console.warn('Got invalid facing: ', this.facing, this);
+      return false;
+    }
+    if (this.displayName !== undefined && (typeof this.displayName !== 'string' || this.displayName.length <= 0 || this.displayName.length > PLAYER_DISPLAY_NAME_MAX_LENGTH)) {
       console.warn('Got invalid displayName: ', this.displayName, this);
       return false;
     }
-    if (typeof this.displayChar !== 'string' || !graphicAsset.haveCharacter(this.displayChar)) {
+    if (this.displayChar !== undefined && (typeof this.displayChar !== 'string' || !graphicAsset.hasCharacter(this.displayChar))) {
```
john0312, line 261: If displayName/displayChar is undefined, I guess we should make it fail?

john0312 (approved):

common/gamelib/player.mjs
ktpss95112: Nope. A normal moving does not involve displayName and displayChar. The information is redundant under many situations.

PR #228, opened 2021-11-06
Fix some minor issues so that the escape-game extension and docker-compose

s3131212:
1. Change the way the escape-game extension handles the terminal server connection. (temporary changes, will revisit here when we decide to use a single Nginx instance.)
2. Allow changing the default uid & gid for the online container. This prevents that `yarn` is executed with root, causing some files in `.yarn` owned by root.


s3131212:
Note:
Use `RUN_DIR="../../hitcon-cat-adventure/run" _UID="$(id -u)" _GID="$(id -g)" docker-compose up` to start.

john0312 (approved):

PR #237, opened 2021-11-07
Update point system to new schema; send notification when points are transferred

s3131212:
As title.

### DIFF
--- a/extensions/point-system/common/client.mjs
+++ b/extensions/point-system/common/client.mjs
@@ -42,7 +42,7 @@ class Client {
   async getPoints() {
     try {
       let ret = await this.requestApi('/users/me', 'GET');
-      return ret?.points;
+      return ret?.data?.points;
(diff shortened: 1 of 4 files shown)


extensions/point-system/common/client.mjs
```diff
@@ -21,7 +21,10 @@ class Client {
   async gameStart() {
     this.helper.mainUI.contextMenu.addToOthersMenu('Transfer Points', 'https://via.placeholder.com/15x15', (player) => {
       // TODO: use modal or something else instead of prompt.
-      this.transferPoints(parseInt(prompt("How many points to transfer:", 0)), player.playerID);
+      if (this.transferPoints(parseInt(prompt("How many points to transfer:", 0)), player.playerID)) {
```
john0312, line 24: Try-catch to handle non-integer case?

john0312 (approved):

PR #245, opened 2021-11-07
Lock player move

john0312:


services/gateway/gateway-service.mjs
```diff
@@ -368,23 +377,27 @@ class GatewayService {
...
-    const ret = await this._enterCoord(msg.mapCoord);
-
-    // If the player was in ghost mode, ignore the occupation check.
-    if (!(socket.playerData.ghostMode || allowOverlap)) {
-      // If the player moves in normal mode, check the occupation.
-      if (!msg.ghostMode && !ret) {
-        await this._leaveCoord(msg.mapCoord);
-        return false;
+    let res;
+    res = await socket.moveLock.acquire('move', async () => {
```
ktpss95112, line 381: `const res = await ...`

services/gateway/gateway-service.mjs
```diff
@@ -210,77 +211,82 @@ class GatewayService {
       }
     }
 
-    // Load the player data.
-    socket.playerData = await this.dir.getPlayerData(playerID);
-    socket.playerID = playerID;
+    // Acquire this lock if you need to move the player.
+    socket.moveLock = new AsyncLock();
+    await socket.moveLock.acquire('move', async () => {
```
ktpss95112, line 216: I think it is overkill to make the whole process locked.
The necessary part, i.e. the code that really deals with moving players, is only `socket.on('avatarSelect', ...`
Note that I suggest a little refactoring in below comment.

services/gateway/gateway-service.mjs
```diff
@@ -210,77 +211,82 @@ class GatewayService {
...
-      this.onPlayerUpdate(socket, PlayerSyncMessage.fromObject(msg));
-      // onPlayerUpdate is async, so returns immediately.
-    });
+      // Synchronize the state.
+      const initLoc = socket.playerData.mapCoord ?? this.gameMap.getRandomSpawnPoint();
+      await this._enterCoord(initLoc);
+      socket.playerData.mapCoord = initLoc;
+      socket.playerData.facing = 'D';
+      socket.playerData.lastMovingTime = Date.now();
+      socket.playerData.ghostMode = false;
```
ktpss95112, line 262: Maybe move these lines into `socket.on('avatarSelect' ...`?

ktpss95112 (commented):
In function `addSocket`, the critical section is only from `this._enterCoord()` to `this._broadcastPlayerUpdate()`. Maybe consider only lock these code instead of almost the whole function.

services/gateway/gateway-service.mjs
john0312: Done.

services/gateway/gateway-service.mjs
john0312: Added comments on why that is done. (To serialize the connect and disconnect handlers.)

services/gateway/gateway-service.mjs
john0312: Done.

john0312:
Reviews have been addressed. There are still some race conditions in this code (should've locked handler for avatarSelect). However, those will be fixed in the coming PR.

ktpss95112 (approved):

PR #246, opened 2021-11-07
Draw map without some layers

lisasasasa:


john0312:
You need to convert both world1 (map01) and world2 (map02) at the same time, into the same `mapsConfigPath`

tiled-maps/scripts/generateMapConfig.mjs
```diff
@@ -210,7 +212,7 @@ function combineSingleLayer(childMaps, layerName) {
       const mapName = `${base}-0${idx}`;
       const targetLayer = childMaps[mapName].layers
         .filter((layer) => layer.name.toLowerCase() === layerName)[0];
-
+      if (targetLayer === undefined) return null;
```
john0312, line 215: The data in combinedLayer might be incorrect, because it's missing the data here in this `mapName`, but it might exist in other `mapName`.

john0312:
Superceded by PR #255

PR #249, opened 2021-11-07
Add haproxy to docker

john0312:


s3131212 (approved):

start-all.mjs
```diff
@@ -116,8 +129,8 @@ async function main() {
 
   /* Start standalone extension services */
   const extServices = {};
-  const enabledExtStandalone = config.get('ext.standalone');
-  for (const ext in enabledExtStandalone) {
+  const enabledExtStandalone = config.get('ext.enabled');
```
s3131212, line 132: I don't think we need to start the standalone extension process if the standalone is empty (e.g. `playerlist`.) This is a rare case, though.

PR #250, opened 2021-11-07
Several Jitsi fixes

s3131212:
1. Maintain a mapping for the Jitsi participant ID and the player ID, in order to hide dangling participants in the Jitsi conference (if a participant ID does not map to any player, hide that participant)
2. Reuse the Jitsi object between connections. This might solve the issue that repeatedly enter and quit Jitsi slow down the browser. 
3. Add a lock to the `unload` function, so that we won't quit a Jitsi conference multiple times.
4. Select the fullscreen participant to get better video streaming quality.

### DIFF
--- a/extensions/jitsi/common/jitsi.mjs
+++ b/extensions/jitsi/common/jitsi.mjs
@@ -429,7 +429,7 @@ class JitsiHandler {
       }
     }
 
-    this.broadcastParticipantId(null); // broadcast that the player has quit.
+    await this.broadcastParticipantId(null); // broadcast that the player has quit.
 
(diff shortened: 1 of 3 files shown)


extensions/jitsi/common/jitsi.mjs
```diff
@@ -412,6 +429,8 @@ class JitsiHandler {
       }
     }
 
+    this.broadcastParticipantId(null); // broadcast that the player has quit.
```
john0312, line 432: Maybe await here or at the end of the function to reduce the chance of race condition?

john0312 (approved):

PR #254, opened 2021-11-13
Overhaul connection and disconnection handler in gateway service

john0312:


john0312:
(Merging without approval due to time constraint. Please let me know if there are any comments. Thanks.)

PR #258, opened 2021-11-14
Fix event delivery to extensions

john0312:


john0312:
Since we've discussed this IRL, I'm merging this first. I'll fix any comments in the following PR.

PR #269, opened 2021-11-14
Fix bug

zeze-zeze:


common/maplib/map.mjs
```diff
@@ -255,6 +255,7 @@ class GameMap {
    * @return {String} cell - The raw content of the cell. '' if any error.
```
john0312, line 255: I think we should return null on error, which is the standard for no value in our map. We'll need to update this comment as well.

john0312 (approved):

john0312 (approved):

PR #274, opened 2021-11-15
jitsi-related UI update

fanlan1210:
* when the focus is not on jitsi overlay, it will become semi-transparent
* update icons of button (screen sharing, fullscreen exit)
* remote jitsi users' video frame will be placed in the scrollable list at the right side of screen


extensions/jitsi/in-div.ejs
```diff
@@ -145,14 +155,27 @@
...
+  <div>
+    <button id="jitsi-others-btn"></button>
+  </div>
 </div>
 
-<div id='jitsi-fullscreen-overlay'></div>
\ No newline at end of file
+<div id='jitsi-fullscreen-overlay'></div>
+
+<div id="jitsi-modal">
```
s3131212, line 176: Not sure what `#jitsi-modal` is for. Others LGTM.

s3131212 (approved):

extensions/jitsi/in-div.ejs
fanlan1210: [TODO]
In the future, it will provide user to view all participants info in the meeting.

canyugs (approved):

PR #279, opened 2021-11-20
Fix layer not found for combineSingleLayer

john0312:


john0312:
(Merging first, please let me know if it doesn't work.)

PR #290, opened 2021-11-21
Convert foreground1 and object layer correctly

john0312:


john0312:
(Merging first, please let me know if there's any comments.)

PR #293, opened 2021-11-21
Better error message for map conversion during character conversion

john0312:


john0312:
(Merging first due to time constraint, please let me know if there's any comments.)

PR #297, opened 2021-11-21
Add map6 and 7 for conversion

john0312:


john0312:
(Merge first due to time constraint.)

PR #300, opened 2021-11-21
Suppress unimportant logs in map conversion

john0312:


john0312:
(Merging first due to time constraint.)

PR #305, opened 2021-11-23
Auto-focusing of input canvas

ktpss95112:
The implementation uses the `tagName` attribute of HTML element.
Not sure if there are some cases missed.

### DIFF
--- a/sites/game-client/input-manager.mjs
+++ b/sites/game-client/input-manager.mjs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: BSD-2-Clause
 
 const KEYSTROKE_RATE = 30; // keystroke per second
+const CANVAS_AUTO_FOCUS_MS = 200;
 
 /**
  * Input manager deals with all user input.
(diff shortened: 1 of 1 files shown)


sites/game-client/input-manager.mjs
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: BSD-2-Clause
 
 const KEYSTROKE_RATE = 30; // keystroke per second
+const CANVAS_AUTO_FOCUS_MS = 200;
```
john0312, line 5: I can't be certain that it doesn't trigger reflow. Maybe double check that it doesn't or I would want to be conservative and put this value at 2000ms or 5000ms to avoid jitter?

john0312 (approved):

sites/game-client/input-manager.mjs
ktpss95112: Can't see any possible code that would trigger a [browser reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a).

PR #310, opened 2021-11-23
Some fixes in the cat adventure extension.

s3131212:
1. Use Set instead of Array to speed up the query to redeemed codes.
2. Remove the finalized check in the quit team function.
3. List of team members now shows the player displayName instead of the playerID.
4. The finalized dialog should be called in the team-manager config, not hardcode in the extension.

### DIFF
--- a/extensions/escape-game/standalone.mjs
+++ b/extensions/escape-game/standalone.mjs
@@ -179,7 +179,7 @@ class Standalone {
    * Quit the team
    * @param {string} playerID The player id.
    */
-   async quitTeam(playerID) {
+  async quitTeam(playerID) {
     // Check if the team exists.
     if (!this.playerToTeam.has(playerID)) {
       throw new Error(`Player ${playerID} trys to quit a non-exist team.`);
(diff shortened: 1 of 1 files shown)


extensions/escape-game/standalone.mjs
```diff
@@ -552,8 +551,7 @@ class Standalone {
 
       this.finalizeTeam(playerID, this.playerToTeam.get(playerID).teamId);
 
-      await this.helper.callS2cAPI(playerID, 'dialog', 'showDialog', 60*1000,
-        sfInfo.name, 'Finalized');
+      // await this.helper.callS2cAPI(playerID, 'dialog', 'showDialog', 60*1000, sfInfo.name, 'Finalized');
```
john0312, line 554: Is this intended?

john0312 (approved):

extensions/escape-game/standalone.mjs
s3131212: Yes, @ccali66 already create a finalized dialog in the team manager. This removes the redundant dialog.

PR #313, opened 2021-11-24
implment help extension

canyugs:


extensions/help-desk/standalone.mjs
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Create the standalone extension service object but does not start it.
+   * @constructor
+   * @param {ExtensionHelper} helper - An extension helper object for servicing
+   * various functionalities of the extension.
+   */
+  constructor(helper) {
+    this.helper = helper;
+    this.locations = {
+      'serviceDesk': new MapCoord('world7', 7, 3),
```
john0312, line 19: If we ever come back to fix this, I would recommend:
```MapCoord.fromSerializedStr(config.get('help-desk.location'))```

john0312 (approved):

PR #330, opened 2021-11-25
Fix a bug in how setting extension handles the checkbox

s3131212:
As title. 

The `elem.value` always return the `value` attribute or `on` for the checkbox. We should use `elem.checked` instead. 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox


s3131212:
By the way, this should fix the bug that changing nickname make the name labels of the players and NPCs disappear.

john0312 (approved):

PR #340, opened 2022-06-20
battle royale extension

Alx-Lai:


extensions/battleroyale/standalone.mjs
```diff
@@ -0,0 +1,338 @@
...
+    this.fireCounter = 0;
+    const updateMap = setInterval( async () => {
+      this.fireCounter++;
+      this.fires.clear();
+      for (const mapName of this.arenaOfMaps.keys()) {
+        const needUpdate = this.updateEdge(
+            this.fires, this.arenaOfMaps.get(mapName), this.fireCounter,
+        );
+        if (!needUpdate) {
+          clearInterval(updateMap);
```
john0312, line 129: We might have to design an auto restart mechanism, such that we'll have multiple rounds running one after another, or players may be able to trigger it to start.

extensions/battleroyale/standalone.mjs
```diff
@@ -0,0 +1,338 @@
...
+     this.mapCoord.y == coord.y;
+  }
+
+  /**
+   * delete the interval
+   * @return {undefined}
+   */
+  deleteInterval() {
+    if (this.intervalID !== undefined) {
+      clearInterval(this.intervalID);
```
john0312, line 62: Maybe clear out this.intervalID as well?

extensions/battleroyale/standalone.mjs
```diff
@@ -0,0 +1,532 @@
...
+              name: LAYER_FIRE.layerName,
+              cells: Array.from(this.fires),
+            }),
+        );
+        const players = this.helper.gameState.getPlayers();
+        players.forEach((player, playerID) => {
+          if (this.helper.gameMap.getCell(player.mapCoord, LAYER_FIRE.layerName)) {
+            this.killPlayer(playerID);
+          }
+        });
```
ktpss95112, line 201: The paramater `playerID` is the index of the array, not the playerID we are familiar with. Make sure that this place uses which one and rename the parameter if needed.

extensions/battleroyale/standalone.mjs
```diff
@@ -0,0 +1,532 @@
...
+        const width = cell.w ?? 1;
+        const height = cell.h ?? 1;
+        if (cell.x <= cellCoord.x && cellCoord.x < cell.x + width &&
+          cell.y <= cellCoord.y && cellCoord.y < cell.y + height) {
+          return true;
+        }
+      }
+    }
+    return false;
+  }
```
ktpss95112, line 228:
```javascript
insideMap(coord) {
  const cellsets = this.arenaOfMaps.get(coord.mapName);
  return cellsets.some((cellset) => cellset.containsMapCoord(coord));
}
```

Anyway, this wheel need not be implemented again.

extensions/battleroyale/standalone.mjs
```diff
@@ -0,0 +1,532 @@
...
+    } else {
+      const players = this.helper.gameState.getPlayers();
+      players.forEach((player, playerID) => {
+        const {x, y} = newBullet.mapCoord;
+        if (player.mapCoord.x == x && player.mapCoord.y == y) {
+          this.killPlayer(playerID);
+          return true;
+        }
+      });
+    }
```
ktpss95112, line 272: no need to `else`

extensions/battleroyale/standalone.mjs
```diff
@@ -0,0 +1,532 @@
...
+
+    // first tick die
+    if (!this.insideMap(newBullet.mapCoord) ||
+    this.helper.gameMap.getCell(newBullet.mapCoord, LAYER_OBSTACLE.layerName)) {
+      return true;
+    } else {
+      const players = this.helper.gameState.getPlayers();
+      players.forEach((player, playerID) => {
+        const {x, y} = newBullet.mapCoord;
+        if (player.mapCoord.x == x && player.mapCoord.y == y) {
```
ktpss95112, line 267: `if (newBullet.mapCoord.equalsTo(player.mapCoord)) {`

extensions/battleroyale/standalone.mjs
```diff
@@ -0,0 +1,532 @@
...
+
+  /**
+   * Client tries to shoot bullet in front(depends on facing) of mapCoord.
+   * The server returns true on success.
+   * @param {Object} player - (TODO: needs docs)
+   * @param {MapCoord} mapCoord - The bullet position.
+   * @param {String} facing - The bullet facing.
+   * @return {Boolean} success or not
+   */
+  async c2s_attack(player, mapCoord, facing) {
```
ktpss95112, line 238: I think that this function only emits the bullet. I think the logic of bullet updating and collision detection should be separated into another function?

ktpss95112: If I understood correctly, the current layout of your function is like

```javascript
async c2s_attack(...) {
  // check if player can attack
  // create bullet
  // check if the bullet hits someone
  setInterval(() => {
    // update bullet
    // check if the bullet hits someone
  });
}
```

, that would have a lot of code duplicated.
Better, you can do this:

```javascript
Bullet.update = (firstTime) => {
  if (!firstTime) {
    // update bullet
  }
  // check if the bullet hits someone
}

async c2s_attack(...) {
  // check if player can attack
  // create bullet
  bullet.update();
  setInterval(bullet.update);
}
```

extensions/battleroyale/standalone.mjs
```diff
@@ -0,0 +1,532 @@
...
+              cells: Array.from(
+                  this.bullets,
+                  ([_, v]) => v['mapCoord'],
+              ),
+            }),
+        );
+      }
+    }, BULLET_SPEED, bulletID);
+
+
```
ktpss95112, line 339: I would suggest that do not use two blank lines inside a block (function).

extensions/battleroyale/standalone.mjs
```diff
@@ -0,0 +1,532 @@
...
+    return true;
+  }
+  /**
```
ktpss95112, line 342: I would suggest that use a line to separate two member functions.

extensions/battleroyale/standalone.mjs
```diff
@@ -0,0 +1,532 @@
...
+      return;
+    }
+    if (mapCoord === 'undefined') {
+      // No location? Just skip.
+      return;
+    }
+
+    const bullets = this.helper.gameMap.getCell(mapCoord, LAYER_BULLET.layerName);
+    if (bullets) {
+      this.killPlayer(playerID);
```
ktpss95112, line 397: Will a player be killed more than once?
If not, please write down the reason somewhere; otherwise, please make sure that this does no harm and write down the clarification somewhere.

extensions/battleroyale/standalone.mjs
```diff
@@ -0,0 +1,542 @@
...
+      return;
+    }
+
+    return;
+  }
+
+
+  // ----------------------------------
+  // State function for NPC
+  // ----------------------------------
```
ktpss95112, line 413: You have some functions below this section but they are not state functions.

extensions/battleroyale/standalone.mjs
ktpss95112: The type of `players` is Map instead of Array. The code is intended.

extensions/battleroyale/standalone.mjs
```diff
@@ -0,0 +1,542 @@
...
+                this.bullets,
+                ([_, v]) => v['mapCoord'],
+            ),
+          }),
+      );
+    }
+    for (const ID of this.updateMapIntervalIDs) {
+      clearInterval(ID);
+    }
+    this.updateMapIntervalIDs = [];
```
ktpss95112, line 538: Just a kind reminder that if `this.updateMapIntervalIDs` is used in other places, you should consider clearing the array instead of assigning a new array since you might want to keep the object referencing the same object.

ktpss95112 (commented):
as the comments show

extensions/battleroyale/standalone.mjs
```diff
@@ -0,0 +1,519 @@
...
+              name: LAYER_FIRE.layerName,
+              cells: Array.from(this.fires),
+            }),
+        );
+        const players = this.helper.gameState.getPlayers();
+        players.forEach(async (player, playerID) => {
+          if (this.helper.gameMap.getCell(player.mapCoord, LAYER_FIRE.layerName)) {
+            await this.killPlayer(playerID);
+          }
+        });
```
ktpss95112, line 199: Just a kind reminder that this piece of code has the same effect as
```javascript
players.forEach((player, playerID) => {
  if (...) this.killPlayer(playerID);
}
```

, since the two versions both create all the promises and not caring about their execution order.

If you want to ensure the execution order, you have to use for loop like this:

```javascript
for (const [playerID, player] of players) {
  if (...) this.killPlayer(playerID);
```

ktpss95112: The same issue happens on line 254 ~ 259.

ktpss95112 (commented):
as the comments show

ktpss95112 (approved):

PR #344, opened 2022-07-10
Hack map generator to work with 2022 map

john0312:


john0312:
Map generator is bad enough that we should rewrite it someday...

john0312:
Merging for now because it's a hack anyway.

PR #345, opened 2022-07-13
Fix a lot of problems in movement-manager

ktpss95112:
From telegram (I'm too lazy):

```
1. 假設有兩個 player update message 被同時送來，理論上應該要 clear timeout，也就是同時只會有一個 pending message，但看起來 clear timeout 沒有在運作
2. _teleportPlayerInternal 如果起始跟結束座標不相同就會失敗（也就是沒有移動的話）
3. 因為 handlePlayerMove 有用到 async await，所以可能會有 context switch 造成的 race condition
```

In short, the original logic of processing playerUpdateMessage is using a cache of size=1 to store the pending playerUpdatemessage. However, this will lead to discarding valid messages when multiple messages flood the server. The refactored code will use an array to store all the pending messages and process them sequentially.

### DIFF
--- a/common/gamelib/move-check.mjs
+++ b/common/gamelib/move-check.mjs
@@ -4,6 +4,9 @@
 const PLAYER_MOVE_TIME_INTERVAL = 100; // ms
 const PLAYER_MOVE_DISTANCE_LIMIT = 1; // cell
 
+// const movementLogger = console;
+const movementLogger = {debug: () => {}};
+
 /**
  * TODO
  * @param {Player} oldPlayerData - TODO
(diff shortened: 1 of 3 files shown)


common/gamelib/move-check.mjs
```diff
@@ -15,6 +15,7 @@ const PLAYER_MOVE_DISTANCE_LIMIT = 1; // cell
 function checkPlayerMove(oldPlayerData, updateMessage, gameMap, clientSide=false) {
   let canMove = true;
   canMove = canMove && _speedCheck(oldPlayerData, updateMessage, clientSide);
+  if (!canMove) console.debug('- speed check failed');
```
john0312, line 18: It seems that we can't easily disable this log message in production, I think we should have something to allow filtering the log level.

john0312 (approved):

PR #346, opened 2022-07-18
a door in the map triggered by the e2s api

aoaaceai:
1. created a layer on the map for doors
2. added an e2s API to trigger the door

### DIFF
--- a/extensions/escape-game/standalone.mjs
+++ b/extensions/escape-game/standalone.mjs
@@ -17,7 +17,8 @@ import {promisify} from 'util';
 import {fileURLToPath} from 'url';
 import InteractiveObjectServerBaseClass from '../../common/interactive-object/server.mjs';
 import {getRunPath, getConfigPath} from '../../common/path-util/path.mjs';
-
+const LAYER_DOOR = {zIndex: 6, layerName: 'escapeGameDoor'};
+import CellSet from '../../common/maplib/cellset.mjs';
 const MAX_PLAYER_PER_TEAM = 5;
 
 // Bring out the FSM_ERROR for easier reference.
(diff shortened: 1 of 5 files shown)


extensions/escape-game/standalone.mjs
```diff
@@ -124,6 +146,33 @@ class Standalone {
     };
   }
 
+  async e2s_openDoor() {
+    var res = await this.helper.varUtils.writeVar("@doorLastOpen", null, null, Date.now().toString());
```
john0312, line 150: You can use a member variable instead of varUtils, because the entire server shares 1 single instance of standalone server for escape-game.

extensions/escape-game/standalone.mjs
```diff
@@ -124,6 +146,33 @@ class Standalone {
...
+      "world1",
+      CellSet.fromObject({
+        name: LAYER_DOOR.layerName,
+        cells: []
+      }),
+    );
+    setTimeout(async function(obj) {
+      var lastOpen = await obj.helper.varUtils.readVar("@doorLastOpen", null, null);
+      lastOpen = parseInt(lastOpen);
+      if (lastOpen + 3000 <= Date.now()) {
```
john0312, line 162: Make this a config? Try this:
```
const config = require('config');
//...
const openTime = config.get('escape-game.doorOpenTime');
if (lastOpen + openTime <= Date.now()) { // ...
```

Then, please remember to adjust /run/config/default.json and /run/config/production.json.
If possible, adjust the corresponding config in hitcon-cat-adventures as well.

extensions/escape-game/standalone.mjs
```diff
@@ -112,6 +113,27 @@ class Standalone {
       v.registerExtStateFunc('showTerminal', 'escape-game', 'sf_showTerminal');
       v.registerExtStateFunc('checkIsInFinalizedTeam', 'escape-game', 'sf_checkIsInFinalizedTeam');
     });
+    this.doorCells = [{'x': 9, 'y': 12, 'h': 1, 'w': 1}];
```
john0312, line 116: If you intend to have only 1 door, then you can store this in the config, see example below. i.e.
```
this.doorCells = [config.get('escape-game.doorPosition')];
```
Otherwise, I would recommend storing the door location as a cellset (seems a little bit elaborate, not sure if we need this.

john0312 (approved):

extensions/escape-game/standalone.mjs
```diff
@@ -146,8 +146,8 @@ class Standalone {
     };
   }
 
-  async e2s_openDoor() {
-    var res = await this.helper.varUtils.writeVar("@doorLastOpen", null, null, Date.now().toString());
+  async e2s_openDoor(duration) {
+    this.doorOpenPressedCount = (this.doorOpenPressedCount === undefined) ? 1 : this.doorOpenPressedCount + 1;
```
john0312, line 150: Consider initializing this in the constructor?

extensions/escape-game/standalone.mjs
```diff
@@ -146,8 +146,8 @@ class Standalone {
     };
   }
 
-  async e2s_openDoor() {
-    var res = await this.helper.varUtils.writeVar("@doorLastOpen", null, null, Date.now().toString());
+  async e2s_openDoor(duration) {
```
john0312, line 149: It would be recommended to check duration's type.

john0312 (approved):

PR #349, opened 2022-07-18
add a answer checker NPC for the player to report the attacker's IP

aoaaceai:
The image is located at [aoaaceai/victim](https://hub.docker.com/r/aoaaceai/victim). A script to generate the victim image will be uploaded to the staging server.


john0312 (commented):
Please make the change to this file instead, thanks~
https://github.com/canyugs/hitcon-cat-adventure/blob/master/run/npc/docker_forensics.json#L48

PR #353, opened 2022-07-25
Add ghost mode button and game button to joystick

ktpss95112:


sites/game-client/body-mobile.ejs
```diff
@@ -61,7 +61,11 @@
           <canvas id="inputcanvas"></canvas>
           <div id="joystick-container">
             <!-- TODO: styling -->
-            <div id="joystick-control-point"></div>
+            <div id="joystick-control-point-container">
+              <div id="joystick-control-point"></div>
+            </div>
+            <div id="joystick-ghost-mode-button" class="container-center hoverable"><img src="/static/sites/game-client/input-manager/ghostmode.png"></div>
+            <div id="joystick-game-button" class="container-center hoverable"><img src="/static/sites/game-client/input-manager/bomb.png"></div>
```
john0312, line 68: Not really sure if we should make this part of the bombman extension, as in, if we remove the bombman extension, the button will disappear. If it takes too much time, we can leave a TODO here instead.

sites/game-client/input-manager/input-manager.mjs
```diff
@@ -319,37 +346,83 @@ class InputManager {
...
-   * @param {Element} containerDiv - The container of the joystick.
+   * @param {InputManager} inputManager - The input manager which this joystick belongs to.
+   * @param {Element} containerDiv - The outer container of the joystick.
+   * @param {Element} controlPointContainerDiv - The container of the joystick control point.
    * @param {Element} controlPointDiv - The div to display the stick of the joystick.
+   * @param {Element} ghostModeDiv - The div to switch on/off ghost mode.
+   * @param {Element} gameButtonDiv - The div used by game (usually by extensions).
    */
-  constructor(containerDiv, controlPointDiv) {
+  constructor(inputManager, containerDiv, controlPointContainerDiv, controlPointDiv, ghostModeDiv, gameButtonDiv) {
```
john0312, line 356: Not really sure if we should hard code the list of buttons we've, but if making it flexible takes too much time, then we can consider adding a TODO here.

john0312 (approved):

sites/game-client/body-mobile.ejs
ktpss95112: The "game button" would be used by any extension which needs a button. It is not limited to bombman extension. For example, battle royale may need to use this button as well.
It requires too much engineering to allow extensions to add buttons. Let it be TODO would be good for now.

sites/game-client/input-manager/input-manager.mjs
ktpss95112: Answered above.

john0312 (approved):
I'm fine with adding a TODO, still LGTM.

PR #354, opened 2022-07-26
add config for new maps

canyugs:


tiled-maps/scripts/generateMapConfig.mjs
```diff
@@ -386,11 +386,23 @@ function loadAndConvertWorld(mapName, mapsDir, tilesetDirectory, resultLayerMap,
...
+  {
+    srcMapName: 'HITCON2022-bombman',
+    dstMapName: 'world2',
+    cellsetsConfig: cellsetsConfig,
+    type: 'single'
+  },
+  {
+    srcMapName: 'HITCON2022-royale',
+    dstMapName: 'world3',
+    cellsetsConfig: cellsetsConfig,
```
john0312, line 403: Consider reading cellsets from another file because we don't want spawn points in every map?

canyugs: Got it, updated

john0312 (approved):

PR #356, opened 2022-07-27
add a c2s function for the player to obtain the certificate

aoaaceai:


extensions/escape-game/standalone.mjs
```diff
@@ -390,6 +390,13 @@ class Standalone {
     }
   }
 
+  /**
+   * The secret function to give a "certificate" item to the player
+   */
+   async c2s_getCertificateItem(player) {
+    return await this.helper.callS2sAPI('items', 'AddItem', player.playerID, 'certificate', 1, 1);
```
john0312, line 397: There are no restrictions on the client side calling this, we should add a flag or secret to make sure only those who've completed the mission get the item. It is recommended that the flag be configurable from config file.

john0312 (approved):

extensions/escape-game/standalone.mjs
```diff
@@ -393,8 +393,13 @@ class Standalone {
   /**
    * The secret function to give a "certificate" item to the player
    */
-   async c2s_getCertificateItem(player) {
-    return await this.helper.callS2sAPI('items', 'AddItem', player.playerID, 'certificate', 1, 1);
+   async c2s_getCertificateItem(player, token) {
+    if (config.has('escape-game.certificateC2sToken') &&
+      config.get('escape-game.certificateC2sToken') == token) {
```
john0312, line 398: Prefer `===` instead of `==`.

extensions/escape-game/standalone.mjs
```diff
@@ -393,8 +393,13 @@ class Standalone {
...
+        return await this.helper.callS2sAPI('items', 'AddItem', player.playerID, 'certificate', 1, 1);
+      } else {
+        return {'ok': false, 'amount': 1};
```
john0312, line 401: I would recommend leaving a message so they know that the token is incorrect.

PR #357, opened 2022-07-27
Add TOTP checking functions for NPC.

john0312:


aoaaceai (approved):

extensions/iobj-lib/standalone.mjs
```diff
@@ -3,11 +3,16 @@
...
 import {randomShuffle, randomChoice} from '../../common/utility/random-tool.mjs';
 import InteractiveObjectServerBaseClass from '../../common/interactive-object/server.mjs';
 import {getRunPath, getConfigPath} from '../../common/path-util/path.mjs';
 
+// Boilerplate for getting require() in es module.
+import {createRequire} from 'module';
+const require = createRequire(import.meta.url);
+
+const otplib = require('otplib');
+
```
ktpss95112, line 15: Is there a way to use `import` instead of `require()`? If yes, then I would suggest not using `require()` to make our code more consistent. (and explicit file path used in `import` is more clear than a mere module name used in `require()` to me)

extensions/iobj-lib/standalone.mjs
```diff
@@ -87,6 +92,27 @@ class Standalone {
...
   }
 
+  /**
+   * Show an open-ended dialog and check if the entered value is
+   * correct, interpreting the entered values as an totp.
+   */
+  async s2s_sf_showDialogAndCheckTOTP(srcExt, playerID, kwargs, sfInfo) {
+    const {nextState, nextStateIncorrect, dialog, secret, otpWindow} = kwargs;
+    const res = await this.helper.callS2cAPI(playerID, 'dialog',
+    'showDialogWithPrompt', 60*1000, sfInfo.visibleName, dialog);
```
ktpss95112, line 102: eslint would tell you that your indentation should be 8 whitespaces.

run/npc/example-npc2.json
```diff
@@ -25,10 +25,21 @@
...
             }
           }
         },
+        "o1": {
+            "func": "showDialogAndCheckTOTP",
+            "kwargs": {
+              "dialog": "Give me your key!",
+              "nextState": "s3",
+              "nextStateIncorrect" : "s4",
+              "secret": "TEST",
```
ktpss95112, line 39: I'm not quite familiar with this implementation. Is `secret` a property of the NPC and might be used in other situations (maybe generating the OTP)? Or is it only used in this state function? If it is used in multiple places, we should define it somewhere else once instead of pasting the credential everywhere where it is needed.

ktpss95112 (commented):
as the comments show

extensions/iobj-lib/standalone.mjs
john0312: I tried import 'otplib' as shown by the otplib documentations, but somehow yarn doesn't like it.

run/npc/example-npc2.json
john0312: Secret is the property of the OTP and this implementation aims to allow multiple OTPs to be checked within the same game, and therefore the OTP secret follows the NPC and not in the config, i.e. different NPC can check different OTP.

OTP generation is outside of this code base so consolidation into config doesn't help.

extensions/iobj-lib/standalone.mjs
john0312: Done.

ktpss95112 (approved):

PR #358, opened 2022-07-30
Fix bug of ghost mode

ktpss95112:
And fix some eslint error.

### DIFF
--- a/services/gateway/gateway-service.mjs
+++ b/services/gateway/gateway-service.mjs
@@ -33,10 +33,12 @@ class GatewayService {
    * @param {Directory} dir - The RPCDirectory for calling other services.
    * @param {GameMap} gameMap - The world map for this game.
    * @param {AuthServer} authServer - Auth server for verifying the token.
+   * @param {AllAreaBroadcaster} broadcaster
    * @param {io} io - Socket.io object.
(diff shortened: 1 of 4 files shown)


services/gateway/gateway-service.mjs
```diff
@@ -499,26 +501,34 @@ class GatewayService {
...
       // If the player doesn't move at all, just return true.
-      if (socket.playerData.mapCoord.equalsTo(msg.mapCoord) && socket.playerData.ghostMode === msg.ghostMode) {
+      if (player.mapCoord.equalsTo(msg.mapCoord) && player.ghostMode === msg.ghostMode) {
         return true;
       }
 
-      const ret = await this._enterCoord(msg.mapCoord);
+      // If the player wants his/her next status is in ghost mode, ignore occupation check and no need to enterCoord().
+      if (msg.ghostMode) {
+        // empty
```
john0312, line 513: If I understand correctly, the logic currently in use is that the player will increase the cell's count whether it is in ghost mode. If you want to change this logic, then maybe you'll need to modify the disconnect logic as well?

ktpss95112: You are right. I will check all the places where redis is accessed.

john0312 (approved):

PR #359, opened 2022-07-31
Add support for checking multiple items

john0312:


extensions/items/standalone.mjs
```diff
@@ -950,6 +967,42 @@ class Standalone {
...
+   *     representing the items.
+   *   kwargs.haveItems specifies the state to go to if the player have the
+   *     amount of specified items.
+   *   kwargs.noItems specifies the state to go to if the player doesn't have
+   *     the amount of specified items.
+   *   kwargs.amount is the minimum total count to go to "haveItems" state.
+   * @return {String} nextState - The next state.
+   */
+  async s2s_sf_haveMultiItems(srcExt, playerID, kwargs, sfInfo) {
+    let amount = kwargs.amount;
```
ktpss95112, line 983: `const amount = kwargs.amount ?? 1;`

extensions/items/standalone.mjs
```diff
@@ -950,6 +967,42 @@ class Standalone {
...
+    let items = kwargs.items;
```
ktpss95112, line 984: `const items = kwargs.items;`

extensions/items/standalone.mjs
```diff
@@ -950,6 +967,42 @@ class Standalone {
...
+
+    if (!Number.isInteger(amount)) amount = 1;
```
ktpss95112, line 986: delete it if above comment resolved

extensions/items/standalone.mjs
```diff
@@ -950,6 +967,42 @@ class Standalone {
...
+
+    let total = 0;
+    for (const itm in result.amounts) {
+      total += result.amounts[itm];
+    }
+    if (total >= amount) {
+      return kwargs.haveItems;
+    } else {
+      return kwargs.noItems;
+    }
```
ktpss95112, line 1003: I think the following code is better since it is more obvious that this function always has a `return` at the end.
```javascript
return (total >= amount) ? kwargs.haveItems : kwargs.noItems;
```

Besides, these two arguments should be explicitly aliased at the beginning of the function so that the developer can see all the required arguments of `kwargs`. If the requirements of `kwargs` is changed in the future, the jsdoc may not be updated, and the new developer will suffer from inconsistency between document and code. My suggestion:
```javascript
async s2s_sf_haveMultiItems(srcExt, playerID, kwargs, sfInfo) {
  const amount = kwargs.amount ?? 1; // I doubt that we shouldn't allow implicit argument. Would be better if asked the developer of NPC to assign the amount everytime
  const {items, haveItems, noItems} = kwargs; // This syntax is so great.
  
  ... // other stuffs
}
```

ktpss95112 (commented):
as the comments show

extensions/items/standalone.mjs
john0312: Done.

extensions/items/standalone.mjs
john0312: Done.

extensions/items/standalone.mjs
john0312: if kwargs.amount is `true`, then the result could be incorrect. I've changed it to use Number.isInteger above.

extensions/items/standalone.mjs
john0312: Done.

ktpss95112 (approved):

PR #372, opened 2022-08-06
Terminal server on demand, done on the server side

aoaaceai:
In contrast to [PR#368](https://github.com/canyugs/hitcon-online/pull/368), this PR cleans up the containers in the terminal server.
If the client hasn't interacted with the container for too long, the server automatically destroys the container and disconnects with the client. This is done by a second-chance algorithm implemented in `terminal-server/main.mjs`.
`terminal-server/config/default.json` can be edited to change the interval.

### DIFF
--- a/extensions/escape-game/standalone.mjs
+++ b/extensions/escape-game/standalone.mjs
@@ -73,7 +73,7 @@ class Standalone {
     fs.readdirSync(getRunPath('terminal')).filter(v => v.endsWith('.json')).forEach((file) => {
       const terminalName = file.slice(0, -('.json'.length));
       const terminal = new TerminalObject(helper, terminalName, getRunPath('terminal', file));
-      this.terminalObjects.set(`iobj-${terminal.objectName}`, terminal);
+      this.terminalObjects.set(`${InteractiveObjectServerBaseClass.IOBJ_PREFIX}${terminal.objectName}`, terminal);
     });
 
     // this.defaultTerminals determines the list of containers to start when the team is created.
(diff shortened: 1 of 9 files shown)


extensions/escape-game/standalone.mjs
```diff
@@ -200,7 +200,6 @@ class Standalone {
     // Create a new team
     let teamId = randomBytes(32).toString('hex');
     this.teams.set(teamId, new Team(teamId, this.terminalServerGrpcService, this.defaultTerminals));
-    this.teams.get(teamId).startContainers(); // do not wait for the container to start.
```
aoaaceai, line 203: Don't need to start the containers on team creation, since it will be done on showTerminal

extensions/escape-game/standalone.mjs
```diff
@@ -510,6 +506,7 @@ class Standalone {
    */
   async s2s_sf_showTerminal(srcExt, playerID, kwargs, sfInfo) {
     await this._finalizeIfNeeded(playerID);
+    await this.playerToTeam.get(playerID).ensureContainerAvailable(sfInfo.visibleName);
```
john0312, line 509: Consider using `sfInfo.name` instead of `sfInfo.visibleName`, the later is only for display.

john0312 (approved):

extensions/escape-game/standalone.mjs
```diff
@@ -957,27 +953,23 @@ class Team {
...
-
-          if (!ret.success) {
-            console.error(`Fail to start container with image ${defaultTerminal.imageName}.`);
-            return false;
-          }
-          this.terminals.set(defaultTerminal.terminalId, ret.containerId);
-        } catch (e) {
-          console.error('Failed to start container: ', e);
-        }
+  async ensureContainerAvailable(terminalId) {
```
john0312, line 956: Please add jsdoc.
Also please comment that all containers will be reaped by default if unused for a certain time. This behaviour is fixed and cannot be changed.

PR #378, opened 2022-08-08
add player disconnect detect

Alx-Lai:


extensions/battleroyale/standalone.mjs
```diff
@@ -367,7 +367,19 @@ class Standalone {
       // No location? Just skip.
       return;
     }
-
+    if (msg.removed) {
```
john0312, line 370: I think we can consider making `killPlayer()` tolerant of player not being online, then change this to call `killPlayer()` instead.

However, it's only a small improvement in terms of reusing code, so it's up to you on if you want to make the change or just merge the current variant as is.

john0312 (approved):

PR #380, opened 2022-08-14
Docker resource limit

aoaaceai:


terminal-server/container-handler.mjs
```diff
@@ -19,6 +20,9 @@ class ContainerHandler {
   constructor(imageName) {
     this.containerName = containerPrefix + randomBytes(32).toString('hex');
     this.imageName = imageName;
+    this.isolatedNetworkName = config.get('isolatedNetworkName');
+    this.memLimit = config.get('memLimit');
+    this.cpuLimit = config.get('cpuLimit');
```
john0312, line 25: Maybe add a TODO to make these optional or specify a default.

john0312 (approved):

PR #382, opened 2022-08-15
change jwtSecret in escape-game to config

anj226:


john0312 (approved):
Please update the production stuff in hitcon-cat-adventure also. Thanks.

